PR #246, opened 2026-04-10
fix(providers): auto-select single LiteLLM integration and preserve existing on update

alezander86:
## Summary

Fixes EPMCDME-11736: LiteLLM project integration was not applied automatically in CLI when exactly one integration was configured. Users were forced to manually select it each time, and previously saved integrations were silently cleared on profile update failures.

## Changes

- **Auto-select single integration**: When exactly one LiteLLM integration is found for the selected project, it is automatically applied — no prompt shown
- **Preserve integration on update**: When updating an existing profile (`isUpdate=true`) and 0 integrations are returned (due to fetch failure or transient API issue), the existing `codeMieIntegration` from the active config is preserved and reused
- **Multiple integrations**: Prompt is still shown when multiple integrations are available (unchanged behavior)
- **Test fix**: Updated `paths.js` mock in `claude.plugin.statusline.test.ts` to include `getCodemieHome`/`getCodemiePath` exports required by `ConfigLoader` static initializer

## Impact

Before: User always sees "Select LiteLLM integration (Optional)" prompt defaulting to "None", even when only one integration exists. A previously saved integration could be cleared after a failed fetch on subsequent `codemie setup` runs.

After: Single integration is auto-selected with a confirmation message `✓ Auto-selected LiteLLM integration: <alias>`. On update with fetch failure, existing integration is reused with `ℹ️ Reusing saved LiteLLM integration: <alias>`.

## Checklist

- [x] Self-reviewed

### DIFF
--- a/src/providers/plugins/sso/sso.setup-steps.ts
+++ b/src/providers/plugins/sso/sso.setup-steps.ts
@@ -168,11 +168,17 @@ export const SSOSetupSteps: ProviderSetupSteps = {
       integrations = [];
     }
 
-    // Always prompt for integration selection
+    // Resolve integration: auto-select if single, prompt if multiple, preserve existing on failure
     let integrationInfo: CodeMieIntegrationInfo | undefined;
 
-    if (integrations.length > 0) {
-      const projectLabel = selectedProject ? ` for project "${selectedProject}"` : '';
+    const projectLabel = selectedProject ? ` for project "${selectedProject}"` : '';
+
+    if (integrations.length === 1) {
+      // Auto-select the only available integration
+      const single = integrations[0];
+      integrationInfo = { id: single.id, alias: single.alias };
+      console.log(chalk.green(`✓ Auto-selected LiteLLM integration: ${chalk.bold(integrationInfo.alias)}\n`));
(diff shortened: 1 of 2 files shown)


src/providers/plugins/sso/sso.setup-steps.ts
```diff
@@ -190,12 +197,25 @@ export const SSOSetupSteps: ProviderSetupSteps = {
...
       integrationInfo = integrationAnswers.integration;
     } else {
-      // Show message if no integrations found
-      const projectLabel = selectedProject ? ` for project "${selectedProject}"` : '';
-      if (integrationsFetchError) {
-        console.log(chalk.dim(`ℹ️  Proceeding without LiteLLM integration (fetch failed)\n`));
-      } else {
-        console.log(chalk.dim(`ℹ️  No LiteLLM integrations configured${projectLabel}\n`));
+      // No integrations found — try to preserve existing one when updating
+      if (isUpdate) {
```
8nevil8: revert this condition. if integration is not found, there is no need to preserve exesting one, since it's no longer valid. api always return integrations if any

PR #257, opened 2026-04-22
feat(skills): add codemie-sdk skill

kostiantyn-pshenychnyi1:
## Checklist

- [ ] Self-reviewed
- [ ] Manual testing performed
- [ ] No breaking changes (or clearly documented)


codemie-ai:

src/cli/commands/sdk/services/assistants.ts
```diff
@@ -0,0 +1,107 @@
...
+    slug: existing.slug ?? "",
+    llm_model_type:
+      params.llm_model_type ?? existing.llm_model_type ?? defaultLlmModel,
+    categories:
+      params.categories ?? existing.categories?.map((c) => c.id) ?? [],
+    toolkits: params.toolkits ?? existing.toolkits ?? [],
+  };
+
+  if (params.temperature) mergedParams.temperature = params.temperature;
+  else if (existing.temperature === null) {
```
BolotnyaBoss: 🔴 bug: `if (params.temperature)` is falsy when value is `0`, which is a valid sampling temperature — the update will silently ignore the caller's intent. Same issue on the `top_p` check below.

Fix both:
```ts
if (params.temperature !== undefined) mergedParams.temperature = params.temperature;
else if (existing.temperature === null) delete mergedParams.temperature;

if (params.top_p !== undefined) mergedParams.top_p = params.top_p;
else if (existing.top_p === null) delete mergedParams.top_p;
```

src/cli/commands/sdk/services/workflows.ts
```diff
@@ -0,0 +1,68 @@
...
+  return client.workflows.get(workflowId);
+}
+
+export async function createWorkflow(
+  client: CodeMieClient,
+  params: WorkflowCreateParams,
+  yamlConfig?: string,
+): Promise<unknown> {
+  const paramsWithDefaults: WorkflowCreateParams = {
+    mode: "Sequential",
```
BolotnyaBoss, line 30: 🔴 bug: spread order breaks the defaults. In JavaScript, `{ mode: "Sequential", ...{ mode: undefined } }` evaluates to `{ mode: undefined }`. So when a caller omits `mode`, `description`, or `shared`, the trailing spread overrides the defaults with `undefined` and the API receives `undefined` for required fields.

Fix — put the fallbacks after the spread:
```ts
const paramsWithDefaults: WorkflowCreateParams = {
  ...(params as Partial<WorkflowCreateParams>),
  mode: params.mode ?? "Sequential",
  description: params.description ?? "",
  shared: params.shared ?? false,
} as WorkflowCreateParams;
```

src/cli/commands/sdk/utils/cli-utils.ts
```diff
@@ -0,0 +1,141 @@
...
+ */
+export async function parseDataInput(
+  dataFlag: string | undefined,
+): Promise<unknown> {
+  if (!dataFlag) {
+    throw new ConfigurationError('No data provided. Use --data \'{"key":"value"}\'');
+  }
+
+  return JSON.parse(dataFlag);
+}
```
BolotnyaBoss, line 35: 🟡 risk: `JSON.parse(dataFlag)` throws a raw `SyntaxError` on malformed input. The error surfaces through `handleSdkError` as a generic `❌ Unexpected token ...` message with no hint that the JSON itself is the problem.

Wrap and rethrow with context:
```ts
try {
  return JSON.parse(dataFlag);
} catch {
  throw new ConfigurationError(`Invalid JSON passed to --data. Check syntax near the beginning of the string.`);
}
```

src/cli/commands/sdk/utils/file-utils.ts
```diff
@@ -0,0 +1,22 @@
...
+ * @param filePaths Array of local file paths
+ * @returns Array of File objects with content and metadata
+ */
+export async function readFilesFromPaths(
+  filePaths: string[],
+): Promise<File[]> {
+  return Promise.all(
+    filePaths.map(async (filePath) => {
+      const content = await readFile(filePath);
+      const name = basename(filePath);
```
BolotnyaBoss, line 21: 🟡 risk: user-supplied file paths are passed directly to `readFile` without validation. The project security policy requires path validation via `src/utils/security.js` (see `validateFilePath` / `PathSecurityError`).

```ts
import { validateFilePath } from "@/utils/security.js";
// inside the map:
validateFilePath(filePath); // throws PathSecurityError on traversal attempts
const content = await readFile(filePath);
```

PR #291, opened 2026-05-11
fix(cli): repair error formatting and harden Windows SSO browser launch

vadimvlasenko:
## Summary

Three regressions reported by a Windows user on v0.1.0 of `codemie assistants chat`:

- **`TypeError: Cannot read properties of undefined (reading 'trim')`** — SDK `ApiError` with empty `message` and detail nested under `response.error.detail` crashed `formatErrorForUser`.
- **`[object Object]`** printed to the console — `logger.error` did `String(error)` on plain context objects.
- **`ENOENT … wsl-utils/index.js`** at module load — broken Nodist install bricked the entire CLI because `open` was imported statically at the top of `sso.auth.ts`, even for commands that never need a browser.

## Fixes

- `src/utils/errors.ts`
  - `extractErrorDetails` now mines `error.response.error.{detail,description,message,help}` for SDK `ApiError` shapes.
  - New `extractMessageFromObject` recurses through nested `error` properties, guarded by a `WeakSet` and depth cap to avoid stack overflow on cycles.
  - `formatErrorForUser` falls back to `error.name` / `'Error'` when message is empty.
  - `wrapText('')` returns `[indent]` instead of `[]` (defensive primitive).
- `src/utils/logger.ts`
  - `logger.error` renders non-`Error` values via `sanitizeLogArgs` + `JSON.stringify` (with `util.inspect` fallback) so context objects no longer collapse to `[object Object]`.
- `src/providers/plugins/sso/sso.auth.ts`
  - `open` import deferred to dynamic `await import('open')` inside `authenticate()`; if the launcher throws (e.g. missing `wsl-utils` on Nodist), the SSO URL is printed for manual completion.
- `src/cli/commands/assistants/chat/index.ts`
  - Interactive chat now passes `{ showToUser: true }` to `handleChatError` so the formatted detail reaches the console (previously file-log only).
- `tests/integration/cli-commands/skills.test.ts`
  - Realigns with current `DO_NOT_TRACK` / `DISABLE_TELEMETRY` behavior for the `add` subcommand (empty string — telemetry runs behind the local egress-guard shim).

## Test plan

- [x] `npm run build` — passes
- [x] `npm run test:unit` — 1613 passed, 1 skipped (90 files)
- [x] Touched integration tests — 12 passed
- [x] New unit tests for empty-message fallback, SDK `response.error.detail` extraction, circular-object guard, structured-object logger output, deferred `open` import, and SSO URL fallback on `open()` failure
- [x] New integration test for end-to-end error reporting through `createErrorContext` + `formatErrorForUser` + `logger.error`
- [ ] Manual sanity check on Windows (Nodist) — recommend confirming with the original reporter before next release tag

## Notes

- `npm run lint` reports 16 pre-existing errors on `main` in `skills-sh-egress-guard.cjs`, `prepare-install-artifacts.mjs`, and `codemie-statusline.mjs`. None are touched by this branch.
- Worth a follow-up version bump so the affected Windows user actually picks this up.

### DIFF
--- a/src/cli/commands/assistants/__tests__/chat-error-output.test.ts
+++ b/src/cli/commands/assistants/__tests__/chat-error-output.test.ts
@@ -1,0 +1,73 @@
+import { afterEach, describe, expect, it, vi } from 'vitest';
+import { ApiError } from 'codemie-sdk';
+
+vi.mock('@/utils/config.js', () => ({
+  ConfigLoader: {
+    load: vi.fn().mockResolvedValue({
+      provider: 'sso',
+      codemieAssistants: [
+        {
+          id: 'assistant-1',
+          name: 'Confluence Expert',
+          slug: 'confluence-expert'
+        }
+      ]
+    })
+  }
+}));
+
+vi.mock('@/utils/auth.js', () => ({
+  getAuthenticatedClient: vi.fn().mockResolvedValue({
+    assistants: {
+      chat: vi.fn().mockRejectedValue(new ApiError('', 500, {
+        error: {
+          detail: 'Confluence credential is missing'
+        }
+      }))
+    }
+  }),
+  promptReauthentication: vi.fn()
+}));
+
+vi.mock('@/utils/logger.js', () => ({
+  logger: {
+    debug: vi.fn(),
+    error: vi.fn()
+  }
+}));
+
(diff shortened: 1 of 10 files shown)


codemie-ai:

PR #298, opened 2026-05-12
fix(proxy): proxy endpoint performance improvements

uladzislausvetlakou:
## Summary

Addresses performance degradation in the local SSO proxy under large inputs and
streaming workloads. The proxy intermediates all LLM API traffic (Claude Code,
Codex, etc.) and had several compounding bottlenecks causing high TTFB on large
prompts, unbounded memory growth during streaming, and degraded throughput with
slow clients.

## Changes

- **Request body streaming**: The proxy no longer unconditionally buffers the
  full POST body before opening an upstream connection. Buffering is now gated
  on whether an active body-transform plugin actually needs it
  (`requiresRequestBody()` interface method). When no transform is needed, the
  raw request stream is piped directly to upstream — eliminating TTFB latency
  proportional to body size.

- **Streaming backpressure**: The main response streaming loop now respects
  `downstream.write()` backpressure — waits for `drain` before pulling the next
  chunk when the write buffer is full. Previously the buffer could grow
  unboundedly with slow clients.

- **Logging plugin memory**: `LoggingInterceptor` was copying every response
  chunk into an unbounded instance-level array, duplicating the entire response
  body in memory. Fixed by: gating collection behind `logger.isDebugMode()`,
  capping buffering at 64 KB, removing a redundant `Buffer.from()` copy per
  chunk, and moving per-request state into a `Map<requestId, state>` to prevent
  leaks on errored requests.

- **DNS lookup cache**: The MCP auth plugin called `dns.lookup()` on every
  relay request for SSRF protection. Results are now cached per hostname with a
  30-second TTL, eliminating repeated cold-DNS latency on the same target.

- **Benchmark script**: Added `scripts/bench-proxy.mjs` — a zero-dependency
  local benchmark (mock upstream, configurable body size and chunk count) to
  measure TTFB, total duration, throughput, and heap usage.

## Impact

Measured locally against a mock upstream (loopback):

| Scenario | TTFB avg | Throughput | Heap |
|---|---|---|---|
| 100 KB body, 50 chunks | 7.3 ms | 13.3 MB/s | 24 MB |
| 500 KB body, 50 chunks | 28.6 ms | 5.4 MB/s | 21 MB (flat) |

Heap remains flat across requests — confirms no per-request memory accumulation
from the logging fix.

## Checklist

- [x] Self-reviewed

### DIFF
--- a/src/providers/plugins/sso/proxy/plugins/claude-request-normalizer.plugin.ts
+++ b/src/providers/plugins/sso/proxy/plugins/claude-request-normalizer.plugin.ts
@@ -137,6 +137,8 @@ class ClaudeRequestNormalizerInterceptor implements ProxyInterceptor {
 
   constructor(private readonly configModel?: string) {}
 
+  requiresRequestBody(): boolean { return true; }
+
   async onRequest(context: ProxyContext): Promise<void> {
     if (!context.requestBody || !context.headers['content-type']?.includes('application/json')) {
       return;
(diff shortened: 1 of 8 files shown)


codemie-ai:

codemie-ai:
[AUTO_CLOSE_WARNING] 🔒 This pull request was automatically closed after being older than 30 days with no activity. Contact maintainers to reopen if needed.

PR #317, opened 2026-05-22
perf(proxy): bound SSE buffering and add backpressure to streaming path

8nevil8:
## Summary

Eliminates per-chunk buffering and microtask overhead in the SSO proxy so Anthropic SSE token streams flow with constant memory and minimal latency.

- **Logging plugin:** Stops accumulating the full response body for streaming (`text/event-stream`) responses. Tracks only bounded head/tail previews (4 KB each) plus chunk/byte counters. For non-streaming JSON, caps captured body at 64 KB.
- **Proxy streaming loop:** Caches the list of interceptors that implement `onResponseChunk`. When none do (the common case), takes a fast path that forwards chunks as-is without the per-chunk hook loop or extra `Buffer.from` copies.
- **Backpressure:** `streamResponse` now awaits `'drain'` when `downstream.write()` returns false, and tears down the upstream when the downstream closes early.

Plan: `docs/superpowers/plans/2026-05-22-sso-proxy-streaming-perf.md`.

## Test Plan

- [x] `npm run build` — clean
- [x] `npm run lint` — zero warnings
- [ ] Smoke test against a real SSO endpoint with a long Claude streaming response: confirm tokens stream visibly, RSS stays roughly constant, debug log contains `[proxy-response]` entry with `mode: 'streaming-bounded'`, `headPreview`, `tailPreview`.
- [ ] Trigger a non-streaming JSON response (e.g. small `/v1/models` call) and confirm the log entry still includes parsed JSON body up to 64 KB, or `[truncated: ...]` for larger bodies.

Generated with AI

Co-Authored-By: codemie-ai <codemie.ai@gmail.com>

### DIFF
--- a/src/providers/plugins/sso/proxy/plugins/logging.plugin.ts
+++ b/src/providers/plugins/sso/proxy/plugins/logging.plugin.ts
@@ -39,16 +39,35 @@ class LoggingInterceptor implements ProxyInterceptor {
   name = 'logging';
   private chunkCount = 0;
   private totalBytes = 0;
-  private responseChunks: Buffer[] = [];
   private responseContentType: string | null = null;
+  private isStreaming = false;
+
+  // Bounded preview buffers — never the full body for SSE.
+  private headPreview: Buffer[] = [];
+  private tailPreview: Buffer[] = [];
+  private headBytes = 0;
+  private tailBytes = 0;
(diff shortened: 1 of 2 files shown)


codemie-ai:

PR #330, opened 2026-06-03
fix(agents): pass --version to downstream binary and suppress banners in non-TTY mode

AlexStasko:
## Summary

In order to use `codemie-*` agents with wrappers like https://github.com/pingdotgg/t3code

## Changes

- `--version` option is passed to the downstream binary, so the upstream tool can understand the real version of the harness.
- suppress banners in non-TTY mode. Mostly for `codemie-claude` binary triggered via Claude SDK

## Checklist

- [x] Self-reviewed

### DIFF
--- a/src/agents/core/AgentCLI.ts
+++ b/src/agents/core/AgentCLI.ts
@@ -62,7 +62,6 @@ export class AgentCLI {
     this.program
       .name(programName)
       .description(`CodeMie ${this.adapter.displayName} - ${this.adapter.description}`)
-      .version(this.version)
       .option('-s, --silent', 'Enable silent mode')
       .option('--status', 'Enable status bar (shows model, context usage, git branch, and cost)')
       .option('--profile <name>', 'Use specific provider profile')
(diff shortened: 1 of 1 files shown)


BolotnyaBoss (commented):
🟡 **risk** — L533: `getVersion()` can throw (exec/network failure) — no try/catch. Unhandled rejection crashes instead of falling back to `this.version`. Wrap in try/catch.

🔴 **bug** — L524: `.version(this.version)` removed from Commander program. Commander's built-in `-V` handler is gone. Manual intercept in `run()` is the only path now — safe only if `run()` is the sole entry point forever. If `program.parseAsync` is ever called directly (tests, future refactor), version breaks silently. Either keep `.version()` on the program and override only the output, or document this hard constraint.

🟡 **risk** — L531: raw argv scan matches `--version`/`-V` anywhere in args — e.g. `codemie run --version` would intercept and exit early instead of forwarding. Guard: only intercept when it's the sole meaningful arg.

🟡 **risk** — L160: `!process.stdout.isTTY` fires in GitHub Actions with `--color` (stdout is a TTY there). Consider also checking `process.env.CI` or `!process.stdin.isTTY && !process.stdout.isTTY` to avoid suppressing banners in CI where they're useful.

codemie-ai:

codemie-ai:
[AUTO_CLOSE_WARNING] 🔒 This pull request was automatically closed after being older than 30 days with no activity. Contact maintainers to reopen if needed.

PR #336, opened 2026-06-05
fix(proxy): sync plugin-bundled skills on proxy startup

AntonYeromin:
## Summary

Skills bundled in the CLI plugin (e.g. `msgraph`) were invisible when using `codemie proxy start` or `codemie proxy connect desktop`, even though the same skills work with `codemie cli`.

Root cause: `--plugin-dir ~/.codemie/claude-plugin` is prepended to Claude Code's args by `sso.template.ts enrichArgs` only in the CLI agent path. In the proxy path that flag is never passed, so Claude Code never loads the plugin skill files.

A previous commit on this branch fixed platform-registered skills (`syncRegisteredSkills`). This commit fixes plugin-bundled skills.

## Changes

- **`src/cli/commands/skills/setup/sync-plugin.ts`** (new) — `syncPluginSkills()` installs the Claude plugin bundle via `ClaudePluginInstaller` and copies each bundled `SKILL.md` to `~/.claude/skills/{name}/SKILL.md` with `${CLAUDE_PLUGIN_ROOT}` resolved to the actual installation path.
- **`src/cli/commands/proxy/index.ts`** — calls `syncPluginSkills().catch(() => {})` fire-and-forget alongside `syncRegisteredSkills()` in both `proxy start` and `proxy connect desktop` startup paths.
- **`src/cli/commands/skills/setup/__tests__/sync-plugin.test.ts`** (new) — unit tests for `syncPluginSkills`.
- **`src/cli/commands/proxy/__tests__/index.test.ts`** — mock and assertions for `syncPluginSkills` in both proxy commands.

## Impact

After `codemie proxy start` or `codemie proxy connect desktop`:
- Plugin-bundled skills (msgraph, codemie-analytics, etc.) are written to `~/.claude/skills/` with resolved absolute paths.
- Skills are available to Claude Code without `--plugin-dir`.

## Checklist

- [x] Self-reviewed
- [x] 5 new/updated tests — all passing (`npm run test:unit`: 105 files, 1755 tests)
- [x] Pre-commit hooks passed (lint, typecheck, secrets scan)
- [x] No breaking changes

### DIFF
--- a/src/cli/commands/proxy/index.ts
+++ b/src/cli/commands/proxy/index.ts
@@ -9,6 +9,7 @@ import {
 } from '../../../utils/errors.js';
 import { logger } from '../../../utils/logger.js';
 import { sanitizeLogArgs } from '../../../utils/security.js';
+import { syncRegisteredSkills } from '../skills/setup/sync.js';
 import {
   checkStatus,
   readState,
@@ -158,6 +159,8 @@ export function createProxyCommand(): Command {
 
       await verifySsoCredentials(config.baseUrl, config.name ?? 'default');
 
+      syncRegisteredSkills(config.name ?? 'default', process.cwd()).catch(() => {});
+
       console.log('Starting proxy daemon...');
       const daemonState = await spawnDaemon({
         targetUrl: config.baseUrl,
@@ -277,6 +280,7 @@ export function createProxyCommand(): Command {
             })
           );
           await verifySsoCredentials(config.baseUrl, config.name ?? 'default');
+          syncRegisteredSkills(config.name ?? 'default', process.cwd()).catch(() => {});
           state = await spawnDaemon({
             targetUrl: config.baseUrl,
             provider: config.provider ?? 'ai-run-sso',
--- a/src/cli/commands/proxy/index.ts
+++ b/src/cli/commands/proxy/index.ts
@@ -160,8 +160,11 @@ export function createProxyCommand(): Command {
 
       await verifySsoCredentials(config.baseUrl, config.name ?? 'default');
 
-      syncRegisteredSkills(config.name ?? 'default', process.cwd()).catch(() => {});
-      syncPluginSkills().catch(() => {});
+      const cwd = process.cwd();
+      await Promise.allSettled([
+        syncRegisteredSkills(config.name ?? 'default', cwd),
+        syncPluginSkills(),
+      ]);
 
       console.log('Starting proxy daemon...');
       const daemonState = await spawnDaemon({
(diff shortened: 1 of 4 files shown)


src/cli/commands/skills/setup/sync-plugin.ts
```diff
@@ -0,0 +1,62 @@
...
+      return;
+    }
+
+    const entries = await fs.readdir(skillsSourceDir, { withFileTypes: true });
+    const claudeSkillsDir = path.join(os.homedir(), '.claude', 'skills');
+
+    for (const entry of entries) {
+      if (!entry.isDirectory()) continue;
+
+      const sourceMd = path.join(skillsSourceDir, entry.name, 'SKILL.md');
```
mykolanehrych: entry.name from fs.readdir goes straight into path.join  if a crafted entry like .. ends up here the read escapes the skills directory. Worth adding a resolve + startsWith guard before the fs.readFile  call.

src/cli/commands/proxy/index.ts
```diff
@@ -158,6 +160,9 @@ export function createProxyCommand(): Command {
 
       await verifySsoCredentials(config.baseUrl, config.name ?? 'default');
 
+      syncRegisteredSkills(config.name ?? 'default', process.cwd()).catch(() => {});
```
mykolanehrych: process.cwd() is not reliable here — detached daemon processes inherit the system's cwd, not the user's project directory. loadSkillsByScope(LOCAL, workingDir) will silently return nothing. The cwd should be captured before the daemon spawns and passed in explicitly.

src/cli/commands/proxy/index.ts
```diff
@@ -277,6 +282,8 @@ export function createProxyCommand(): Command {
             })
           );
           await verifySsoCredentials(config.baseUrl, config.name ?? 'default');
+          syncRegisteredSkills(config.name ?? 'default', process.cwd()).catch(() => {});
```
mykolanehrych: process.cwd() is not reliable here — detached daemon processes inherit the system's cwd, not the user's project directory. loadSkillsByScope(LOCAL, workingDir) will silently return nothing. The cwd should be captured before the daemon spawns and passed in explicitly.

src/cli/commands/proxy/index.ts
```diff
@@ -158,6 +160,9 @@ export function createProxyCommand(): Command {
 
       await verifySsoCredentials(config.baseUrl, config.name ?? 'default');
 
+      syncRegisteredSkills(config.name ?? 'default', process.cwd()).catch(() => {});
+      syncPluginSkills().catch(() => {});
+
       console.log('Starting proxy daemon...');
       const daemonState = await spawnDaemon({
```
mykolanehrych, line 170: Both syncs above are fire-and-forget — daemon can start before they finish writing. Consider await Promise.allSettled([syncRegisteredSkills(...), syncPluginSkills()]) here instead of the .catch(() => {})pattern above.

mykolanehrych:
Approved

PR #339, opened 2026-06-08
fix(config): prevent migration 004/005 from creating empty config on fresh installs

mykolanehrych:
## Summary

- **Jira:** https://jiraeu.epam.com/browse/EPMCDME-12693
- **Root cause**: Migration 004 called `ConfigLoader.loadMultiProviderConfig()` unconditionally. On fresh installs (e.g. Docker containers) with no global config, it returns an empty skeleton `{version:2, profiles:{}}`, which migration then saves to disk. On next startup, `ConfigLoader.load()` finds `profiles: {}` and throws `'No profiles configured'`.
- **Fix 1** (004): wrap global config processing in `if (hasGlobal)` — skips global part when no config exists, but still processes local project config if present.
- **Fix 2** (005): apply the same `hasGlobalConfig` guard for consistency — prevents 005 from running against an empty skeleton on fresh installs.

## Code review findings

4 findings were surfaced during review. 2 fixed in this PR, 2 deferred.

### Fixed in this PR
| # | File | Finding |
|---|------|---------|
| 1 | `004-skills-assistants-top-level.migration.ts` | Early return skipped local config migration when no global config existed — replaced with conditional block |
| 2 | `005-skill-slug-format.migration.ts` | No `hasGlobalConfig` guard — added for consistency with 004 |

### Deferred (separate ticket)
| # | File | Finding |
|---|------|---------|
| 3 | `runner.ts` | Runner permanently records skipped migrations (`migrated:false`) as applied — migration never re-runs after user creates config post-install |
| 4 | `utils/config.ts` | `hasGlobalConfig` reads the config file, then `loadMultiProviderConfig` reads it again — redundant double I/O with TOCTOU window |

## Test plan

- [ ] Fresh Docker container with `--jwt-token`: no `codemie-cli.config.json` created, no `'No profiles configured'` error
- [ ] Existing install with global config: migration 004/005 still runs and migrates skills/assistants to top-level as before
- [ ] Project with local config only (no global): local config is still migrated by 004/005
- [ ] Both global and local config present: both are migrated as before

### DIFF
--- a/src/migrations/004-skills-assistants-top-level.migration.ts
+++ b/src/migrations/004-skills-assistants-top-level.migration.ts
@@ -12,13 +12,13 @@ class SkillsAssistantsTopLevelMigration implements Migration {
     let migrated = false;
 
     const hasGlobal = await ConfigLoader.hasGlobalConfig();
-    if (!hasGlobal) return { success: true, migrated: false, reason: 'no-existing-config' };
-
-    const globalConfig = await ConfigLoader.loadMultiProviderConfig();
-    const migratedGlobal = this.migrate(globalConfig);
-    if (migratedGlobal !== globalConfig) {
-      await ConfigLoader.saveMultiProviderConfig(migratedGlobal);
-      migrated = true;
+    if (hasGlobal) {
+      const globalConfig = await ConfigLoader.loadMultiProviderConfig();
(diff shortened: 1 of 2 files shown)


Dark-Sun (approved):

Dark-Sun:
approved

PR #346, opened 2026-06-10
fix(cli): add "code" bin entry for npx package resolution

AntonYeromin:
## Summary
Adds a `\"code\"` entry to the `bin` field in `package.json` so that `npx @codemieai/code` can resolve which executable to run. Previously, npm/npx would fail with \"could not determine executable to run\" (or fall back to looking for `code` in PATH) because the package exposes multiple binaries but none matched the unscoped package name `code`.

## Changes
- Added `\"code\": \"./bin/codemie.js\"` to the `bin` field in `package.json`

## Impact
- Before: `npx @codemieai/code setup` → `npm error could not determine executable to run` / `sh: code: command not found`
- After: `npx @codemieai/code setup` works as documented in the README

## Checklist
- [x] Self-reviewed
- [x] No breaking changes

### DIFF
--- a/package.json
+++ b/package.json
@@ -7,6 +7,7 @@
   "types": "dist/index.d.ts",
   "bin": {
     "codemie": "./bin/codemie.js",
+    "code": "./bin/codemie.js",
     "codemie-code": "./bin/agent-executor.js",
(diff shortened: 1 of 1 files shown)


mykolanehrych:
approved

PR #347, opened 2026-06-10
fix(agents): fix reasoning model routing for gpt-5.4 and gpt-5.5

vladyslav-yurchenko1:
## Summary

Adds gpt-5.5 model support with correct routing to OpenAI Responses API and consolidates Responses API model detection into a single source of truth.

## Changes

- Added `gpt-5.5` and `gpt-5.5-2026-04-24` static entries with `use_responses_api: true` in `opencode-model-configs.ts`
- Moved `RESPONSES_API_MODEL_PATTERNS` and `isResponsesApiModel()` from `opencode-dynamic-models.ts` to `opencode-model-configs.ts` (single source of truth)
- Fixed `getModelConfig()` fallback to apply `isResponsesApiModel()` so unknown future models are auto-routed correctly
- Removed duplicate pattern array from `opencode-dynamic-models.ts`

## Impact

Before: gpt-5.5 was routed to `/v1/chat/completions` — no reasoning output, tool calls crashed with `reasoning_effort not supported` error.
After: gpt-5.5 routes to `/v1/responses` — reasoning blocks visible in CLI, tool calls work correctly.

## Checklist

- [x] Self-reviewed

### DIFF
--- a/src/agents/plugins/opencode/opencode-model-configs.ts
+++ b/src/agents/plugins/opencode/opencode-model-configs.ts
@@ -228,6 +228,66 @@ export const OPENCODE_MODEL_CONFIGS: Record<string, OpenCodeModelConfig> = {
     }
   },
 
+  'gpt-5.4': {
+    id: 'gpt-5.4',
+    name: 'GPT-5.4',
+    displayName: 'GPT-5.4',
+    family: 'gpt-5',
+    tool_call: true,
+    reasoning: true,
+    attachment: true,
+    temperature: false,
+    structured_output: true,
+    modalities: {
+      input: ['text', 'image'],
+      output: ['text']
+    },
+    knowledge: '2025-08-31',
+    release_date: '2026-03-05',
+    last_updated: '2026-03-05',
+    open_weights: false,
+    use_responses_api: true,
+    cost: {
+      input: 1.75,
+      output: 14,
+      cache_read: 0.175
+    },
+    limit: {
+      context: 400000,
+      output: 128000
+    }
+  },
+
+  'gpt-5.4-2026-03-05': {
+    id: 'gpt-5.4-2026-03-05',
+    name: 'GPT-5.4 (Mar 2026)',
+    displayName: 'GPT-5.4 (Mar 2026)',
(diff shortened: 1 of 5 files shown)


src/providers/plugins/sso/proxy/plugins/request-sanitizer.plugin.ts
```diff
@@ -61,8 +70,13 @@ class RequestSanitizerInterceptor implements ProxyInterceptor {
       const bodyStr = context.requestBody.toString('utf-8');
       const body = JSON.parse(bodyStr);
 
+      // Responses API (/v1/responses) supports reasoningSummary and must not have it stripped.
+      // Chat Completions (/v1/chat/completions) and any other path use the restricted set.
+      const isResponsesApi = context.url?.includes('/v1/responses') || context.url === '/responses';
```
mykolanehrych: context.url is assigned directly from req.url — the /v1 prefix is ​​never stripped, so the second condition (=== '/responses') is dead code. The first condition (includes) is a substring match and would 
incorrectly treat /v1/responses/stream as a Responses API request. Fix: context.url === '/v1/responses'

src/agents/plugins/opencode/opencode-model-configs.ts
```diff
@@ -228,6 +263,126 @@ export const OPENCODE_MODEL_CONFIGS: Record<string, OpenCodeModelConfig> = {
     }
   },
 
+  'gpt-5.4': {
```
mykolanehrych, line 304: 'gpt-5.4': { cost: { input: 1.75, output: 14, cache_read: 0.175 }, ... }
  'gpt-5.4-2026-03-05': { cost: { input: 1.75, output: 14, cache_read: 0.175 }, ... }

  These four config blocks are nearly identical  only id, name, and displayName differ. If pricing changes, it needs to be updated in multiple places and it's easy to miss one. maybe worth extracting shared fields into a base object?

mykolanehrych:
Approved

codemie-ai (approved):

PR #375, opened 2026-06-17
fix(agents): correctly import Claude Desktop file-analysis conversations

EvgeniyKvasyuk:
## Claude Desktop file-analysis import fix

Claude Desktop (Cowork) conversations that include a file upload and analysis were imported into CodeMie incompletely:

- the assistant's analysis reply was missing (empty response);
- the uploaded file appeared as raw `<uploaded_files>` XML in both the message and the chat name, and opening it returned a 500;
- the conversation showed up twice — the real chat plus an empty duplicate.

With this change such conversations import fully and correctly: the user prompt, the attached file name, and the complete assistant reply are shown in a single chat.

Verified end-to-end against a real Cowork file-analysis session; unit tests cover the message and file transformation.


Dark-Sun:
approved

Dark-Sun:
@AntonYeromin

PR #384, opened 2026-06-23
feat(analytics): add OTEL analytics source with session-detail report

vadimvlasenko:
## Summary

Adds an `analytics otel --file <path>` source that maps a flattened native
Claude Code OTEL events file (`otel-events.jsonl`) into the existing analytics
pipeline, rendering the same HTML report (cost, tokens, tools, timeline, skills,
turns, project/branch) from OTEL data. The default local-session source is
unchanged.

## Changes

- **AnalyticsSource seam** (`sources/{types,sessions-source,otel-source}.ts`):
  the CLI is now agnostic to where data comes from; `index.ts` is split into a
  shared `runAnalytics` runner plus the `otel` subcommand.
- **`otel-loader.ts`**: parse/filter OTEL events, authoritative cost index
  (native `cost_usd`), per-model rollup, timeline dispatches, synthesized
  `MetricDelta`s, and a transcript cwd/branch join. Each agent `api_request` is
  attributed to a single dispatch, so parallel subagents with overlapping
  windows do not double-count cost/tokens.
- **`otel-source.ts`**: honors `--session/--project/--agent/--branch` so
  filtered runs return the right sessions and reports are never mislabeled.
- **`usage-readers.ts`**: keeps the most complete row when Claude writes
  progressive JSONL records for a single streaming response.
- **`hook.ts`**: on SessionStart re-entry, preserves accumulated
  `startTime`/`activeDurationMs` for a *live* session (`compact`); `clear`/exit
  start fresh instead of resurrecting a completed session.
- **report** (`app.js`, `template.html`): prompt/title column, per-session JSON
  export, full-window timeline.
- Test fixtures use **anonymized** user identifiers (no real emails/usernames).

## Testing

- [x] Tests added/updated (otel-loader unit + otel-report integration;
  usage-readers progressive-row case)
- [x] Manual testing done (`analytics otel --file <fixture> --report` renders
  HTML + JSON; filter and dispatch attribution verified end to end)

## Checklist

- [x] CI is green (`CI=true npm run ci` — 2106 unit + 211 integration passing)

> Note: `cost-enricher.test.ts` "TTL-aware pricing against real transcripts"
> reads machine-specific local transcripts and is `it.skip`ped when `CI` is set
> (pre-existing behavior, unrelated to this change).


BolotnyaBoss:
The architecture seam is clean, the dispatch attribution logic works correctly for the core case, and the plan-vs-implementation is fully aligned.

PR #387, opened 2026-06-23
fix(config): surface codemieAssistants and codemieSkills through ConfigLoader

8nevil8:
## Summary

`ConfigLoader.loadGlobalConfigProfile()` and `loadLocalConfigProfile()` spread only per-profile
data when returning a resolved profile, silently dropping `codemieAssistants` and `codemieSkills`
which live at the `MultiProviderConfig` root (not inside any profile). This caused the Registered
Assistants tab to always appear empty even after running `codemie setup brianna` or similar.

## Changes

- **`src/utils/config.ts`**: both `loadGlobalConfigProfile` and `loadLocalConfigProfile` now
  forward `rawConfig.codemieAssistants` and `rawConfig.codemieSkills` from the root into the
  returned profile object
- **`src/env/types.ts`**: added missing `codemieSkills?: CodemieSkill[]` to `ProviderProfile`
  so the field is type-safe through the full `CodeMieConfigOptions` chain
- **`src/utils/__tests__/config-project-override.test.ts`**: four new tests covering global
  assistants/skills forwarding and the "local without field does not overwrite global" invariant
- **`src/cli/commands/analytics/cost/__tests__/cost-enricher.test.ts`**: pre-existing acceptance
  tests that required local transcript files now skip gracefully when those files are absent
  (replaced `itLocal` with a per-file `itSession` factory using `existsSync`)

## Testing

- [x] 4 new unit tests covering the fix
- [x] All 2095 unit tests pass, 220 integration tests pass
- [x] Manual: `codemie setup brianna` → `Registered Assistants` tab now shows the assistant

## Checklist

### DIFF
--- a/docs/superpowers/tasks/2026-06-23-registered-assistants-empty-tab/complexity-assessment.json
+++ b/docs/superpowers/tasks/2026-06-23-registered-assistants-empty-tab/complexity-assessment.json
@@ -1,0 +1,30 @@
+{
+  "schema": 1,
+  "task": "Fix registered assistants tab showing empty by surfacing codemieAssistants (and codemieSkills) from the MultiProviderConfig top level through ConfigLoader.load().",
+  "generated": "2026-06-23T00:00:00Z",
+  "dimensions": {
+    "component_scope":      { "score": 2, "label": "S" },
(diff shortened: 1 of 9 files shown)


codemie-ai:

PR #413, opened 2026-07-09
fix(cli): report correct version in success message after install

SleepySML:
## Summary

- Fixes incorrect version shown in success message after \`codemie install claude <version>\` when a previous version is already installed
- Changes \`AgentAdapter.installVersion()\` return type from \`Promise<void>\` to \`Promise<string | null>\` so the verified installed version propagates to the caller instead of being discarded
- Extracts \`execVersionAtFullPath()\` private helper in \`ClaudePlugin\` — shared by \`getVersion()\` and \`isInstalled()\` to eliminate asymmetry; removes the incorrect \`shell: true\` (ELF binary executes directly; \`shell: true\` breaks paths with spaces)
- Adds version mismatch warning in \`install.ts\`: if the detected version doesn't match the requested one (Windows PATH-refresh scenario), shows an honest warning instead of asserting a wrong version

## Changes

- \`src/agents/core/types.ts\` — \`installVersion?()\` return type: \`Promise<void>\` → \`Promise<string | null>\`
- \`src/agents/core/BaseAgentAdapter.ts\` — wrap \`getVersion()\` in try/catch after \`npm.installGlobal()\`; return \`null\` on error instead of propagating (binary may not yet be on PATH post-install)
- \`src/agents/plugins/claude/claude.plugin.ts\` — extract private \`execVersionAtFullPath()\` used by both \`getVersion()\` and \`isInstalled()\`; return \`result.installedVersion ?? null\` from \`installVersion()\`
- \`src/agents/plugins/kimi/kimi.plugin.ts\` — return \`await this.getVersion()\` after \`this.installNative()\`
- \`src/cli/commands/install.ts\` — capture return value of \`installVersion()\`; use it for success message; fall back to \`getVersion()\` when null; guard version mismatch check with semver regex so channel aliases (\`latest\`, \`stable\`, \`supported\`) never trigger the restart warning
- Updated tests in \`kimi.plugin.test.ts\` and \`install.version-selection.test.ts\`
- \`.ai-run/guides/architecture/architecture.md\` — add \`installVersion?()\` to \`AgentAdapter\` interface reference
- \`.ai-run/guides/development/development-practices.md\` — document defensive try-catch pattern for post-install \`getVersion()\` and semver-guard heuristic
- \`docs/superpowers/tasks/…/spec.md\` — annotate \`shell: true\` section with decision-reversal note

## Testing

- \`npm test\` passes (all 4 tests in \`install.version-selection.test.ts\` including new mismatch scenario)
- E2E verified in WSL with \`~/.local/bin\` not in PATH (full-path verification only):
  - Downgrade 2.1.205 → 2.1.33: \`Claude Code v2.1.33 installed successfully\` ✅
  - Upgrade 2.1.33 → 2.1.34: \`Claude Code v2.1.34 installed successfully\`, \`claude --version\` confirms \`2.1.34\` ✅
- When detected version mismatches requested (Windows stale PATH): \`spinner.warn()\` with honest message instead of wrong \`spinner.succeed()\`
- Channel alias installs (\`codemie install claude latest\`) no longer produce a spurious restart warning
- \`update\` and \`setup\` flows unaffected (they ignore the return value)

## Checklist

- [x] Tests updated and passing
- [x] No regressions in \`update\`/\`setup\` flows
- [x] Reviewer feedback addressed (CR-001: try-catch on getVersion; CR-002: semver guard on mismatch check)
- [x] Knowledge guides updated (\`architecture.md\`, \`development-practices.md\`, \`spec.md\`)
- [x] Conventional Commits format used
- [x] Linked to [EPMCDME-10318](https://jiraeu.epam.com/browse/EPMCDME-10318)


src/agents/core/BaseAgentAdapter.ts
```diff
@@ -178,6 +178,8 @@ export abstract class BaseAgentAdapter implements AgentAdapter {
       }
       throw error;
     }
+
+    return await this.getVersion();
```
yanaSelin: **CR-001 — Unguarded `getVersion()` after successful install**

If `getVersion()` throws here (subprocess error, permission denied, binary not yet on PATH), the exception propagates to `install.ts` and the user sees an install-failure error even though the npm package was installed successfully.

```suggestion
    try {
      return await this.getVersion();
    } catch {
      return null;
    }
```

`install.ts` already handles a `null` return (falls back to a fresh `getVersion()` call or omits the version from the success message), so returning `null` on error is the safe path.

src/cli/commands/install.ts
```diff
@@ -174,32 +174,50 @@ export function createInstallCommand(): Command {
...
+            const installedVersionStr = displayVersion ? ` v${displayVersion}` : '';
+
+            // Sanity-check: if verification found a version that doesn't match what was requested,
+            // the installer ran but PATH still resolves the old binary (common on Windows after install).
+            // Report honestly instead of asserting the wrong version as success.
+            const requestedVersion = actualVersionToInstall;
+            const versionMismatch =
+              requestedVersion &&
+              displayVersion &&
+              displayVersion !== requestedVersion;
```
yanaSelin, line 199: **CR-002 — Mismatch check fires on channel aliases, producing spurious restart warning**

`requestedVersion` here is `actualVersionToInstall`, which can be a channel alias (`'latest'`, `'stable'`, `'supported'`). The comparison `'latest' !== '2.1.34'` is always true, so every channel-based install (`codemie install claude latest`) triggers the "A terminal restart may be required" warning even when the install succeeded exactly as expected.

```suggestion
            const versionMismatch =
              requestedVersion &&
              /^\d+\.\d+\.\d+/.test(requestedVersion) &&
              displayVersion &&
              displayVersion !== requestedVersion;
```

Only compare when `requestedVersion` is a concrete semver — channel aliases should never fail the mismatch check.

yanaSelin (changes requested):
Two major issues found in the version-reporting fix. Both are straightforward patches.

**CR-001** — `BaseAgentAdapter`: the new `return await this.getVersion()` is unguarded. If `getVersion()` throws after a successful npm install, the exception propagates to the CLI and the user sees an install-failure error even though the package was installed. Wrap it in `try/catch { return null; }`.

**CR-002** — `install.ts`: the `versionMismatch` guard compares `actualVersionToInstall` against `displayVersion` without first confirming it is a semver string. When a user runs `codemie install claude latest` (or `stable`, `supported`), `actualVersionToInstall` is the alias string; `'latest' !== '2.1.34'` is always true, so every channel-based install triggers the "terminal restart may be required" warning regardless of outcome. Guard with `/^\d+\.\d+\.\d+/.test(requestedVersion)` before comparing.

src/agents/core/BaseAgentAdapter.ts
SleepySML: Done — addressed in the latest commit.

src/cli/commands/install.ts
SleepySML: Done — addressed in the latest commit.

src/cli/commands/__tests__/install.version-selection.test.ts
```diff
@@ -74,4 +76,109 @@ describe('install command version selection', () => {
...
+    const { createInstallCommand } = await import('../install.js');
+    const command = createInstallCommand();
+
+    await command.parseAsync(['node', 'codemie', 'claude', '2.1.34']);
+
+    expect(spinnerSucceedMock).not.toHaveBeenCalled();
+    expect(spinnerWarnMock).toHaveBeenCalledWith(
+      expect.stringContaining('v2.1.33') &&
+      expect.stringContaining('v2.1.34') &&
+      expect.stringContaining('terminal restart')
```
yanaSelin: JavaScript `&&` short-circuits and evaluates to the last operand. `expect.stringContaining('v2.1.33') && expect.stringContaining('v2.1.34') && expect.stringContaining('terminal restart')` resolves to just `expect.stringContaining('terminal restart')` — the first two checks are silently dropped. Use three separate assertions:

```ts
const [actualArg] = spinnerWarnMock.mock.calls[0];
expect(actualArg).toContain('v2.1.33');
expect(actualArg).toContain('v2.1.34');
expect(actualArg).toContain('terminal restart');
```

src/agents/plugins/claude/claude.plugin.ts
```diff
@@ -316,6 +316,18 @@ export class ClaudePlugin extends BaseAgentAdapter {
    *
    * @returns Version string or null if not installed
    */
+  private async execVersionAtFullPath(): Promise<string | null> {
+    if (process.platform === 'win32') return null;
+    const { exec } = await import('../../../utils/processes.js');
+    const fullPath = resolveHomeDir('.local/bin/claude');
+    try {
+      const result = await exec(fullPath, ['--version']);
+      return result.code === 0 ? result.stdout.trim() : null;
```
yanaSelin: If the binary exits with code 0 but produces no stdout, `result.stdout.trim()` returns `""`. An empty string is not `null`, so `getVersion()` enters the `if (fullPathOutput !== null)` branch, the regex match fails, and it returns `""` instead of falling through to the PATH-based check. `isInstalled()` would also return `true` for empty output.

Guard against this:

```ts
const trimmed = result.stdout.trim();
return trimmed.length > 0 ? trimmed : null;
```

docs/superpowers/tasks/2026-07-08-epmcdme-10318-fix-version-reporting/spec.md
```diff
@@ -0,0 +1,93 @@
...
+```
+
+### Per-file changes
+
+| File | Change |
+|---|---|
+| `src/agents/core/types.ts` | `installVersion?()` return type: `Promise<void>` → `Promise<string \| null>` |
+| `src/agents/core/BaseAgentAdapter.ts` | Return `await this.getVersion()` after `npm.installGlobal()` |
+| `src/agents/plugins/claude/claude.plugin.ts` | Return `result.installedVersion ?? null` from `installVersion()`; add `shell: true` to primary exec in `getVersion()` |
+| `src/agents/plugins/kimi/kimi.plugin.ts` | Return `await this.getVersion()` after `this.installNative()` |
```
yanaSelin: The per-file change table says the Kimi plugin should "Return `await this.getVersion()` after `this.installNative()`", but the actual implementation returns `await this.installNative(resolvedVersion)`, which propagates `result.installedVersion ?? null` directly (correct and better — avoids a redundant shell exec). Update the row to match what was actually built.

yanaSelin (changes requested):

src/agents/plugins/claude/claude.plugin.ts
SleepySML: Done — guarded against empty stdout: `result.stdout.trim()` is now assigned to `trimmed`, and we return `null` if `trimmed.length === 0`. Addressed in the latest commit.

src/cli/commands/__tests__/install.version-selection.test.ts
SleepySML: Done — replaced the `&&`-chained `stringContaining` with three separate `expect(actualArg).toContain()` assertions, addressed in the latest commit.

docs/superpowers/tasks/2026-07-08-epmcdme-10318-fix-version-reporting/spec.md
SleepySML: Done — updated the Kimi row to reflect that `installVersion()` delegates to `installNative()` and returns `result.installedVersion ?? null` directly (no redundant `getVersion()` call), addressed in the latest commit.

yanaSelin (approved):
All review comments resolved correctly:

- **CR-001**: try-catch on post-install getVersion() — correct, matches suggestion
- **CR-002**: semver guard on mismatch check — correct, channel aliases no longer trigger warning
- **Empty stdout guard** in execVersionAtFullPath — correct, returns null for zero-length output
- **Split test assertions** — three separate toContain() calls, no silent dropping
- **Spec.md Kimi row** — updated to match actual implementation

No new issues found. shell:true correctly removed from execVersionAtFullPath in the follow-up commit. Type signatures consistent throughout.

PR #414, opened 2026-07-09
chore(agents): bump Claude Code to 2.1.205 and Codex CLI to 0.143.0

homich1991:
## Summary
- Bump Claude Code supported version: `2.1.199` → `2.1.205`, minimum: `2.1.189` → `2.1.195`
- Bump Codex CLI supported version: `0.129.0` → `0.143.0`, minimum: `0.119.0` → `0.133.0`
- Update Codex version-support tests to match new versions
- Update Claude client integration tests

## Test Plan
- [x] `vitest run` passes (113/113 tests)
- [x] TypeScript typecheck clean
- [x] ESLint zero warnings


scripts/postinstall.mjs
```diff
@@ -1,8 +1,20 @@
...
+import { fileURLToPath } from 'node:url';
+
+// npm strips execute bits on macOS when unpacking prebuilt binaries
+if (process.platform === 'darwin') {
+  const root = join(dirname(fileURLToPath(import.meta.url)), '..');
+  for (const arch of ['darwin-arm64', 'darwin-x64']) {
+    const helper = join(root, 'node_modules', 'node-pty', 'prebuilds', arch, 'spawn-helper');
+    if (existsSync(helper)) {
+      try { chmodSync(helper, 0o755); } catch { /* non-fatal: may not be writable */ }
+    }
```
yanaSelin, line 15: **CR-001 — silent chmod failure, no user diagnostic**

If `chmodSync` fails here (binary owned by another user, read-only FS, permission denied), postinstall exits cleanly — but `spawn-helper` remains non-executable and PTY sessions will fail later with a cryptic error and no actionable guidance.

At minimum, surface the failure:
```js
try { chmodSync(helper, 0o755); } catch (err) {
  process.stderr.write(`[codemie] Warning: could not set execute bit on ${helper} — PTY sessions may fail.\n  Reason: ${err.message}\n`);
}
```

tests/setup/agent-build-setup.ts
```diff
@@ -31,6 +31,16 @@ let originalSsoProfile: string | undefined;
  * Ensures dist/ exists and the claude CLI is installed before agent tests run.
  */
 export async function setup(): Promise<void> {
+  // npm strips execute bits on macOS when unpacking prebuilt node-pty binaries
+  if (process.platform === 'darwin') {
+    for (const arch of ['darwin-arm64', 'darwin-x64']) {
+      const helper = join(root, 'node_modules', 'node-pty', 'prebuilds', arch, 'spawn-helper');
+      if (existsSync(helper)) {
+        try { chmodSync(helper, 0o755); } catch { /* non-fatal */ }
+      }
```
yanaSelin, line 40: **CR-001 (same as postinstall) — catch swallows failure silently**

Same issue as in `scripts/postinstall.mjs`: if `chmodSync` fails in the test setup, integration tests will fail with a cryptic PTY error instead of an actionable message. Add a `console.warn` in the catch.

src/agents/plugins/codex/__tests__/codex.plugin.version-support.test.ts
```diff
@@ -82,7 +82,7 @@ describe('CodexPlugin version support', () => {
 
     expect(compat.installedVersion).toBe('0.118.9');
     expect(compat.isBelowMinimum).toBe(true);
-    expect(compat.minimumSupportedVersion).toBe('0.119.0');
+    expect(compat.minimumSupportedVersion).toBe('0.133.0');
   });
```
yanaSelin, line 86: **CR-002 — `isBelowMinimum` test is no longer a boundary test**

The test above uses `0.118.9` as the installed version, which was previously just below the old minimum (`0.119.0`, off by `0.0.1`). With the new minimum at `0.133.0`, `0.118.9` is 14 minor versions below — no longer exercising the boundary.

The actual boundary (`0.132.9` should be blocked; `0.133.0` should not) is untested. Consider adding:
```ts
// just below new minimum — must be blocked
vi.mocked(processes.exec).mockResolvedValueOnce({ code: 0, stdout: 'codex-cli 0.132.9\n', stderr: '' });
const plugin2 = new CodexPlugin();
const compat2 = await plugin2.checkVersionCompatibility();
expect(compat2.isBelowMinimum).toBe(true);
```

yanaSelin (changes requested):
Two findings — both straightforward to fix before merge.

**CR-001** (major): `chmodSync` failure is silently swallowed in two places; if it fails, postinstall exits clean but PTY sessions break later with no diagnostic.

**CR-002** (major): the `isBelowMinimum` test no longer exercises the new minimum boundary — it uses `0.118.9` against `0.133.0`, which is far from the edge. The boundary at `0.132.9` vs `0.133.0` is unverified.

homich1991 (commented):
Done — addressed in the latest commit (e2c08a9).

CR-001: both `chmodSync` catch blocks now emit `console.warn` with the helper path and error message instead of swallowing silently.
CR-002: `isBelowMinimum` test updated to use `0.132.9` (one patch below the new minimum `0.133.0`) to verify the actual boundary condition.

8nevil8 (approved):

PR #424, opened 2026-07-15
fix(config): fix duplicate model in /model picker for haiku/opus-only tenants (EPMCDME-12779)

Sergei-Nikitin-epam:
## Summary

- **Root cause**: \`autoSelectModelTiers()\` set \`sonnetModel\` only when \`selectedModel.includes('sonnet')\`. Selecting haiku as primary left \`sonnetModel\` undefined even when the tenant had real sonnet models, so \`ANTHROPIC_DEFAULT_SONNET_MODEL\` was never set and Claude Code fell back to the haiku model ID for the Custom Sonnet slot — producing a duplicate entry in \`/model\`.
- **Fix**: Apply the same models-list auto-selection for the sonnet tier as already done for haiku and opus: filter by the \`'sonnet'\` keyword and pick the latest version regardless of which tier was chosen as primary.
- **Also re-introduces** the opus-only and haiku-only tenant routing that was reverted in #423, with the haiku-only case added:
  - **Opus-only**: set only \`CLAUDE_CODE_SUBAGENT_MODEL\` (not \`ANTHROPIC_DEFAULT_SONNET_MODEL\`) to prevent opus appearing in both Custom Sonnet and Custom Opus slots.
  - **Haiku-only**: route haiku through the sonnet slot and omit \`ANTHROPIC_DEFAULT_HAIKU_MODEL\` to prevent haiku appearing in both slots.

## Files changed

| File | Change |
|---|---|
| \`src/cli/commands/setup.ts\` | \`autoSelectModelTiers()\` — sonnet now auto-selects from models list like haiku/opus |
| \`src/agents/core/BaseAgentAdapter.ts\` | \`transformEnvVars()\` — opus-only and haiku-only slot routing |
| \`src/providers/plugins/bedrock/bedrock.template.ts\` | Bedrock hook — same opus-only / haiku-only fallback branches |
| \`src/cli/commands/__tests__/model-tier-auto-selection.test.ts\` | New tests: haiku-primary-with-sonnet and no-sonnet-in-list cases |
| \`src/agents/core/__tests__/model-tier-config.test.ts\` | Updated haiku-only test to reflect sonnet-slot routing |

## Test plan

- [x] \`npm run test -- model-tier-auto-selection\` — new EPMCDME-12779 test cases pass
- [x] \`npm run test -- model-tier-config\` — haiku-only routing test passes
- [x] Manual: run \`codemie setup\` with haiku as primary model on a tenant that also provisions sonnet → \`/model\` picker shows correct distinct IDs in Custom Haiku and Custom Sonnet slots
- [x] Manual: opus-only tenant → no duplicate in Custom Sonnet / Custom Opus slots
- [x] Manual: haiku-only tenant → model appears once (in sonnet slot), no duplicate

Closes EPMCDME-12779

### DIFF
--- a/src/agents/core/BaseAgentAdapter.ts
+++ b/src/agents/core/BaseAgentAdapter.ts
@@ -1062,16 +1062,7 @@ export abstract class BaseAgentAdapter implements AgentAdapter {
     }
 
     // Transform model tiers (haiku/sonnet/opus)
-    if (envMapping.haikuModel) {
-      for (const envVar of envMapping.haikuModel) delete env[envVar];
-    }
-    if (envMapping.sonnetModel) {
-      for (const envVar of envMapping.sonnetModel) delete env[envVar];
-    }
-    if (envMapping.opusModel) {
-      for (const envVar of envMapping.opusModel) delete env[envVar];
-    }
-
+    // Note: All tier vars were already cleared in Step 1 above
     if (env.CODEMIE_HAIKU_MODEL && envMapping.haikuModel) {
       for (const envVar of envMapping.haikuModel) {
         env[envVar] = env.CODEMIE_HAIKU_MODEL;
--- a/src/agents/core/BaseAgentAdapter.ts
+++ b/src/agents/core/BaseAgentAdapter.ts
@@ -1062,15 +1062,36 @@ export abstract class BaseAgentAdapter implements AgentAdapter {
     }
 
     // Transform model tiers (haiku/sonnet/opus)
+    if (envMapping.haikuModel) {
+      for (const envVar of envMapping.haikuModel) delete env[envVar];
+    }
+    if (envMapping.sonnetModel) {
+      for (const envVar of envMapping.sonnetModel) delete env[envVar];
+    }
+    if (envMapping.opusModel) {
+      for (const envVar of envMapping.opusModel) delete env[envVar];
+    }
+
     if (env.CODEMIE_HAIKU_MODEL && envMapping.haikuModel) {
       for (const envVar of envMapping.haikuModel) {
         env[envVar] = env.CODEMIE_HAIKU_MODEL;
       }
     }
-    if (env.CODEMIE_SONNET_MODEL && envMapping.sonnetModel) {
+    if (env.CODEMIE_SONNET_MODEL && env.CODEMIE_SONNET_MODEL !== env.CODEMIE_HAIKU_MODEL && envMapping.sonnetModel) {
+      // Distinct sonnet tier — map to all target vars normally
       for (const envVar of envMapping.sonnetModel) {
         env[envVar] = env.CODEMIE_SONNET_MODEL;
       }
+    } else if ((!env.CODEMIE_SONNET_MODEL || env.CODEMIE_SONNET_MODEL === env.CODEMIE_HAIKU_MODEL) && env.CODEMIE_OPUS_MODEL && envMapping.sonnetModel?.includes('CLAUDE_CODE_SUBAGENT_MODEL')) {
+      // No distinct sonnet tier, opus provisioned: route subagent to opus.
+      // ANTHROPIC_DEFAULT_SONNET_MODEL is intentionally left unset to prevent
+      // duplicate-ID display in /model (EPMCDME-12779).
+      env['CLAUDE_CODE_SUBAGENT_MODEL'] = env.CODEMIE_OPUS_MODEL;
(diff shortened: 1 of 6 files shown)


src/agents/core/BaseAgentAdapter.ts
```diff
@@ -1062,15 +1062,36 @@ export abstract class BaseAgentAdapter implements AgentAdapter {
...
     // Transform model tiers (haiku/sonnet/opus)
+    if (envMapping.haikuModel) {
+      for (const envVar of envMapping.haikuModel) delete env[envVar];
+    }
+    if (envMapping.sonnetModel) {
+      for (const envVar of envMapping.sonnetModel) delete env[envVar];
+    }
+    if (envMapping.opusModel) {
+      for (const envVar of envMapping.opusModel) delete env[envVar];
+    }
```
yanaSelin: **[CR-004 MAJOR] Dead code — these delete blocks are already executed in Step 1**

Step 1 of this method (around line 1024-1040 in the same file) unconditionally clears `haikuModel`, `sonnetModel`, and `opusModel` env vars before any transformation runs. The three new blocks here (lines 1065-1073) fire after Step 1 and delete the same vars a second time — they can never remove anything that Step 1 hasn't already cleared.

This won't cause a runtime bug today, but it misleads future maintainers into thinking the tier-routing block is self-contained, which it isn't — it depends on Step 1 for the initial clear. If Step 1 is ever narrowed or moved, the two clearing passes can silently drift out of sync.

Suggestion: Remove these three delete blocks. If the goal is to make the tier section self-contained, relocate the Step 1 clearing to just before line 1064 and remove it from Step 1.

yanaSelin (changes requested):
## Code Review — 3 findings (2 critical, 1 major)

The fix correctly addresses the duplicate-model-in-/model-picker symptom in the common case, but two bugs in the surrounding code mean the fix doesn't work for existing users who already have a saved profile or a populated shell environment.

---

### CR-001 [CRITICAL] — `src/cli/commands/setup.ts:299-301` — Stale tier config survives re-setup

```typescript
// setup.ts lines 299-301 (NOT changed by this PR)
if (modelTiers.haikuModel) config.haikuModel = modelTiers.haikuModel;
if (modelTiers.sonnetModel) config.sonnetModel = modelTiers.sonnetModel;
if (modelTiers.opusModel) config.opusModel = modelTiers.opusModel;
```

When `autoSelectModelTiers()` returns `{ sonnetModel: undefined }` (e.g. an opus-only tenant), the `if (modelTiers.sonnetModel)` guard is falsy, so the assignment is skipped and **the existing `config.sonnetModel` from the saved profile is left unchanged**. `exportProviderEnvVars` then emits it as `CODEMIE_SONNET_MODEL = 'claude-sonnet-4-6'`, and `transformEnvVars` hits the "distinct sonnet tier" branch — the new opus-only routing never applies.

Any user re-running `codemie setup` to switch to an opus-only or haiku-only tenant will still see the duplicate.

**Fix:** Remove the `if` guards; assign unconditionally:
```typescript
config.haikuModel = modelTiers.haikuModel;
config.sonnetModel = modelTiers.sonnetModel;
config.opusModel = modelTiers.opusModel;
```

---

### CR-002 [CRITICAL] — `src/cli/commands/setup.ts:587-597` — Early-return bypasses new auto-selection logic when shell has stale env vars

```typescript
// setup.ts lines 587-597 (NOT changed by this PR)
if (envHaiku && envSonnet && envOpus) {
  return { haikuModel: envHaiku, sonnetModel: envSonnet, opusModel: envOpus };
}
```

`autoSelectModelTiers()` reads `ANTHROPIC_DEFAULT_HAIKU_MODEL`, `ANTHROPIC_DEFAULT_SONNET_MODEL`, and `ANTHROPIC_DEFAULT_OPUS_MODEL` directly from the process environment. If all three are set in the shell — which is the normal state after any prior multi-tier session — the function returns immediately with the old values. The new `else if (sonnetModels.length > 0)` and `else { // no sonnet tier }` branches are never reached. The `config.ts` empty-string override only covers `CODEMIE_*` vars, not the `ANTHROPIC_DEFAULT_*` vars that this early-return reads.

This means the fix is entirely bypassed for users who have run `codemie setup` before.

**Fix:** Either remove the early-return, or validate that the env vars are consistent with the `models` list before trusting them (e.g. skip `envSonnet` if `models` contains no sonnet-keyword entry).

src/agents/core/BaseAgentAdapter.ts
Sergei-Nikitin-epam: All findings have been addressed.

yanaSelin (approved):

PR #425, opened 2026-07-16
fix(agents): detect and warn on ANTHROPIC_BASE_URL override in ~/.claude/settings.json

alex-budanov:
## Summary

Fixes EPMCDME-10988. \`codemie-code\` injects \`ANTHROPIC_BASE_URL\` via the active profile, but Claude Code silently overrides that value if \`~/.claude/settings.json\` also contains \`ANTHROPIC_BASE_URL\`. The startup banner showed the profile URL while the session used a different endpoint — invisible to the user.

This change detects the override at startup and prints a visible warning showing both URLs before the session starts. It also detects \`ANTHROPIC_MODEL\` overrides (scope expanded from the original spec — approved).

## Changes

- **\`src/agents/plugins/claude/settings-conflict.ts\`** (new) — pure async detection helper; reads \`~/.claude/settings.json\`, returns \`ConflictInfo | null\` (covers \`ANTHROPIC_BASE_URL\` and \`ANTHROPIC_MODEL\`)
- **\`src/agents/plugins/claude/claude.plugin.ts\`** — calls \`detectSettingsConflict\` in \`beforeRun\`; prints \`chalk.yellow\` warning to stderr; wrapped in \`try/catch\` so a detection failure never aborts startup; sanitizes URL and model values against ANSI/terminal injection
  - Security hardening: added C1-form OSC \`\x9d\` to DCS pre-strip regex (CR-001); clears \`url.hash\` in \`safeUrl\` (CR-002); wraps \`conflict.profileModel\` in \`safeUrl()\` before output (CR-003)
- **\`src/agents/plugins/claude/__tests__/settings-conflict.test.ts\`** (new) — 16 unit tests covering all edge cases including empty-string URL, filesystem errors, and model conflict detection
- **\`src/agents/plugins/claude/__tests__/claude.plugin.conflict.test.ts\`** (new) — 17 integration tests covering warning emission, empty-string fallback, try/catch degradation, ANSI stripping, and regression guards for CR-001/CR-002/CR-003 injection vectors
- **\`.ai-run/guides/usage/project-config.md\`** — documents \`~/.claude/settings.json\` precedence and how to resolve it
- **\`vitest.config.ts\`** — \`hookTimeout\` increased 10s → 30s (WSL2 cold-start on \`/mnt/c\`)

## Testing

**Automated (33 tests):**
- [x] \`npx vitest run src/agents/plugins/claude/__tests__/settings-conflict.test.ts\` — 16 passed
- [x] \`npx vitest run src/agents/plugins/claude/__tests__/claude.plugin.conflict.test.ts\` — 17 passed
- [x] \`npm run typecheck\` — clean
- [x] \`npm run build\` — clean

**Manual E2E smoke test:**

Since \`codemie-code\` is not yet in production and the full CLI is interactive, the warning was verified by importing \`ClaudePluginMetadata.lifecycle.beforeRun\` directly from the built \`dist/\` in a minimal Node.js script:

1. Built the project (\`npm install && npm run build\`)
2. Injected a conflicting URL into \`~/.claude/settings.json\` via Python one-liner
3. Ran a \`smoke.mjs\` script that calls \`beforeRun\` with the profile URL in \`env\` — no network calls, exits in under a second
4. Confirmed the ⚠️ warning appeared with the correct profile URL and active URL
5. Updated \`~/.claude/settings.json\` to match the profile URL, re-ran — confirmed no warning
6. Restored \`~/.claude/settings.json\` to its original state

## Checklist

- [x] Tests added (33 tests)
- [x] No secrets or unsafe logging
- [x] Architecture boundaries respected (Plugin layer only)
- [x] Documentation updated

Closes EPMCDME-10988

🤖 Generated with [Claude Code](https://claude.ai/claude-code)


src/agents/plugins/claude/settings-conflict.ts
```diff
@@ -0,0 +1,33 @@
...
+
+  let settings: Record<string, unknown>;
+  try {
+    const raw = await readFile(settingsPath, 'utf-8');
+    settings = JSON.parse(raw) as Record<string, unknown>;
+  } catch {
+    return null;
+  }
+
+  const settingsUrl = settings.ANTHROPIC_BASE_URL;
```
yanaSelin: **CR-001 [CRITICAL] — Detection never fires: wrong key path in `settings.json`**

`settings.ANTHROPIC_BASE_URL` is always `undefined`. Claude Code stores env vars under a nested `env` object, so the correct path is `settings.env.ANTHROPIC_BASE_URL`.

As-is, `detectSettingsConflict` always returns `null` regardless of the file contents — the feature is entirely non-functional and no warning is ever shown.

**Fix:**
```typescript
const envBlock = settings.env as Record<string, unknown> | undefined;
const settingsUrl = envBlock?.ANTHROPIC_BASE_URL;
if (typeof settingsUrl !== 'string' || !settingsUrl) return null;
```

The unit test mocks also need updating — see the comment on `settings-conflict.test.ts`.

src/agents/plugins/claude/__tests__/settings-conflict.test.ts
```diff
@@ -0,0 +1,116 @@
...
+    vi.mocked(fsp.readFile).mockResolvedValue(JSON.stringify({ statusLine: 'some-value' }) as any);
+
+    const result = await detectSettingsConflict({ ANTHROPIC_BASE_URL: PROFILE_URL });
+
+    expect(result).toBeNull();
+  });
+
+  it('returns null when settings.json ANTHROPIC_BASE_URL equals env value', async () => {
+    vi.mocked(fsMod.existsSync).mockReturnValue(true);
+    vi.mocked(fsp.readFile).mockResolvedValue(JSON.stringify({ ANTHROPIC_BASE_URL: PROFILE_URL }) as any);
```
yanaSelin: **CR-001 [CRITICAL] — Test mock reflects the wrong `settings.json` schema**

This (and every subsequent `mockResolvedValue(JSON.stringify({ ANTHROPIC_BASE_URL: ... }))` in this file) puts the key at the root of the object, but Claude Code stores env vars under a nested `env` block.

All mocks need to become:
```typescript
JSON.stringify({ env: { ANTHROPIC_BASE_URL: PROFILE_URL } })
```

Because the implementation also reads from root (which is wrong), these tests currently pass while validating broken behaviour — they will start correctly catching the bug only after the implementation fix.

src/agents/plugins/claude/claude.plugin.ts
```diff
@@ -221,6 +221,40 @@ export const ClaudePluginMetadata: AgentMetadata = {
...
+      // Claude Code reads settings.json at startup and silently overrides env vars;
+      // warn the user so they know which endpoint is actually in use.
+      try {
+        const { detectSettingsConflict } = await import('./settings-conflict.js');
+        const conflict = await detectSettingsConflict(env);
+        if (conflict) {
+          // Strip C0/C1 control characters and ANSI CSI sequences before displaying
+          // URL values from settings.json — prevents terminal injection via crafted URLs.
+          const safeUrl = (s: string): string =>
+            s.replace(/[\x00-\x1f\x7f]|\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]/g, '');
```
yanaSelin: **CR-002 [MAJOR] — Dead regex alternative: CSI bracket residue leaks into output**

`\x1b` (0x1B = 27) falls within `[\x00-\x1f]`, so the first alternative consumes the ESC byte before the engine can attempt the full CSI pattern. A crafted URL like `https://proxy/\x1b[31mFORGED\x1b[0m` is sanitized to `https://proxy/[31mFORGED[0m` — ESC is stripped (live ANSI injection is prevented) but the bracket-sequence residue stays and appears as literal visible text in the terminal warning.

**Fix** — swap order so the full CSI sequence is consumed atomically first:
```typescript
const safeUrl = (s: string): string =>
  s.replace(/\x1b\[[\x30-\x3f]*[\x20-\x2f]*[\x40-\x7e]|[\x00-\x1f\x7f]/g, '');
```

src/agents/plugins/claude/settings-conflict.ts
```diff
@@ -0,0 +1,33 @@
...
+import { readFile } from 'fs/promises';
+import { join } from 'path';
+import { resolveHomeDir } from '../../../utils/paths.js';
+
+export interface ConflictInfo {
+  settingsUrl: string;
+  profileUrl: string | undefined;
+}
+
+export async function detectSettingsConflict(
```
yanaSelin, line 13: **CR-003 [MAJOR — decision needed] — `ANTHROPIC_MODEL` override not detected**

Acceptance criterion 3 requires displaying "the actual endpoint/**model** being used." This function only reads `ANTHROPIC_BASE_URL`; if `settings.json` also overrides `ANTHROPIC_MODEL`, the session silently uses a different model than the profile specifies — the same invisible-override problem this ticket was opened to fix, left unaddressed for the model dimension.

The spec's *Out of Scope* section explicitly excludes other keys, creating a gap between the original AC and the delivered scope.

**Decision required:**
- **In scope for this PR:** extend `ConflictInfo` with optional `settingsModel`/`profileModel` fields, read `settings.env.ANTHROPIC_MODEL` here, and add a model row to the chalk warning block when a model conflict is detected.
- **Deliberately deferred:** update AC-3 in the work item to document the narrowed scope so the gap is tracked.

yanaSelin (changes requested):
**Code review — EPMCDME-10988 (3 findings: 1 critical, 2 major)**

All three review lenses ran (blind, edge-case, acceptance). Two findings are straight patches; one needs a scope decision from the product owner before it can be resolved.

See inline comments for details.

src/agents/plugins/claude/settings-conflict.ts
alex-budanov: Done — addressed in the latest commit. Fixed: read from `settings.env.ANTHROPIC_BASE_URL` (the actual Claude Code schema) and updated all test mocks to use `{ env: { ANTHROPIC_BASE_URL: ... } }` to match the real `settings.json` structure.

src/agents/plugins/claude/__tests__/settings-conflict.test.ts
alex-budanov: Done — addressed in the latest commit. All mocks updated to `{ env: { ANTHROPIC_BASE_URL: ... } }`. These now validate the correct schema, so a future regression in the key path will fail the tests.

src/agents/plugins/claude/claude.plugin.ts
alex-budanov: Done — swapped alternation order so the full CSI sequence (`\x1b[...`) is consumed atomically before the C0 catch-all. Added a new test that injects `\x1b[31mFORGED\x1b[0m` and asserts neither `[31m` nor `[0m` residue appears in output.

src/agents/plugins/claude/settings-conflict.ts
alex-budanov: Deliberately deferred — the spec's *Out of Scope* section explicitly excludes other env keys. Leaving this as a tracked gap; a follow-up ticket should extend `ConflictInfo` with `settingsModel`/`profileModel` and add the model row to the warning block.

src/agents/plugins/claude/settings-conflict.ts
alex-budanov: Follow-up (latest commit): added docs fix — project-config.md now shows the correct `env` block JSON snippet. Also added a regression test asserting a root-level `ANTHROPIC_BASE_URL` key (no `env` nesting) returns null.

src/agents/plugins/claude/claude.plugin.ts
alex-budanov: Follow-up (latest commit): upgraded beyond the alternation-order fix. Now using `strip-ansi` (already a direct dep at 7.1.2) plus a C1 control-range blocklist `[\\x00-\\x1f\\x7f-\\x9f]`. This also covers the single-byte C1 CSI form (`\\x9b`) that the hand-rolled regex missed. New test added for `\\x9b`.

src/agents/plugins/claude/settings-conflict.ts
alex-budanov: Update: CR-003 (JSON null guard) is now fixed in the latest commit. Added `if (!settings || typeof settings !== 'object' || Array.isArray(settings)) return null;` after JSON.parse to guard null/array values. Tests for JSON null and JSON array inputs both confirmed RED before the fix and pass GREEN after.

alex-budanov:
**Second-pass fixes (CR-001 / CR-002 / CR-003 from tech lead review)**

All three items addressed via TDD (test written RED, confirmed fail, then implementation to GREEN):

| Finding | Fix |
|---------|-----|
| **CR-001 docs** | `project-config.md` now shows the correct `env` block JSON snippet |
| **CR-001 regression** | Added test: root-level `ANTHROPIC_BASE_URL` (no `env` nesting) → `null` |
| **CR-002** | `safeUrl` now uses `strip-ansi` + C1 blocklist `[\x00-\x1f\x7f-\x9f]`; covers single-byte C1 CSI (`\x9b`) the hand-rolled regex missed |
| **CR-003** | Added `!settings \|\| typeof settings !== 'object' \|\| Array.isArray(settings)` guard after `JSON.parse`; tests for JSON null and JSON array inputs |

19 tests passing, no lint/typecheck warnings.

src/agents/plugins/claude/claude.plugin.ts
```diff
@@ -221,6 +222,67 @@ export const ClaudePluginMetadata: AgentMetadata = {
...
+          const safeUrl = (s: string): string => {
+            // ESC-form: P=DCS X=SOS ^=PM _=APC; C1-form: \x90 \x98 \x9e \x9f
+            const noStringCmds = s.replace(/(?:\x1b[PX^_]|[\x90\x98\x9e\x9f])[\s\S]*?(?:\x07|\x1b\\|\x9c|$)/g, ''); // eslint-disable-line no-control-regex
+            const stripped = stripAnsi(noStringCmds).replace(/[^\x20-\x7e]/gu, '');
+            try {
+              const url = new URL(stripped);
+              if (url.username || url.password) {
+                url.username = '';
+                url.password = '';
+                return `[credentials removed] ${url.toString()}`;
```
yanaSelin, line 254: **CR-001 [major / security]: `safeUrl` does not strip query parameters**

`url.username` and `url.password` are cleared, but `url.searchParams` is not touched. If `ANTHROPIC_BASE_URL` contains credentials in query-string form — e.g. `https://proxy.example.com/?api_key=s3cr3t` — they pass through `safeUrl` and are printed verbatim to `console.error`, where CI log aggregators can capture them.

The project convention is explicit: raw secrets/tokens must not appear in logs (`sanitizeLogArgs` policy in security guide).

Suggested fix — also clear `url.search` in the same branch:
```typescript
if (url.username || url.password) {
  url.username = '';
  url.password = '';
  url.search = '';          // ← add this: strip query params too
  return `[credentials removed] ${url.toString()}`;
}
```
Alternatively, rebuild the display URL from `protocol + host + port + pathname` only, discarding both userinfo and query entirely.

src/agents/plugins/claude/settings-conflict.ts
```diff
@@ -0,0 +1,35 @@
+import { existsSync } from 'fs';
+import { readFile } from 'fs/promises';
+import { join } from 'path';
+import { resolveHomeDir } from '../../../utils/paths.js';
+
+export interface ConflictInfo {
+  settingsUrl: string;
+  profileUrl: string | undefined;
+}
```
yanaSelin, line 11: **CR-002 [critical / decision needed]: AC-3 is only partially satisfied — model identifier is never displayed**

Story AC-3 requires: *"The CLI displays the actual endpoint/**model** being used."*

`ConflictInfo` exposes `settingsUrl` and `profileUrl` only — there are no fields for the active model. As a result:
- The warning block shows the active endpoint, but never the model identifier.
- `ANTHROPIC_MODEL` overrides in `~/.claude/settings.json` produce no warning at all.
- The display is conditional on a conflict being present — there is no always-on startup banner for the no-conflict case.

Extend `ConflictInfo` with `settingsModel` / `profileModel` fields, read `settings.env.ANTHROPIC_MODEL` alongside `ANTHROPIC_BASE_URL`, and include the model in the warning block.

src/agents/plugins/claude/claude.plugin.ts
alex-budanov: Done — url.search is now cleared alongside url.username/url.password, and the condition also fires when query params are present independently (no userinfo required). Commit: e7f4d8a.

src/agents/plugins/claude/settings-conflict.ts
alex-budanov: Going with Option B. Extended `ConflictInfo` with optional `settingsModel`/`profileModel` fields. `detectSettingsConflict` now reads `settings.env.ANTHROPIC_MODEL` alongside `ANTHROPIC_BASE_URL` and returns both when either conflicts. The warning block conditionally renders a model row when a model conflict is present. 5 new unit tests (16 total). Commit: 03cf5e5.

alex-budanov:
## Security hardening follow-up (commit 8317768)

Three terminal-injection gaps identified during code review were patched in a follow-up commit:

**CR-001 — OSC C1-form `\x9d` not stripped**
The DCS pre-strip regex covered ESC-form and most C1-form introducers but omitted `\x9d` (single-byte C1 OSC). A crafted `settingsUrl` containing `\x9d0;payload\x07` would have its control byte stripped but the payload text leaked as printable ASCII. Fixed by adding `\x9d` to the character class.

**CR-002 — `url.hash` not cleared in `safeUrl`**
`url.username`, `url.password`, and `url.search` were explicitly cleared, but `url.hash` was not. A fragment like `#injected-section` survives both ANSI strip and the ASCII allowlist and renders verbatim in the warning output. Fixed by adding `url.hash = ''` to the clearing block.

**CR-003 — `conflict.profileModel` not sanitized before terminal output**
`settingsUrl`, `profileUrl`, and `settingsModel` are all passed through `safeUrl()` before `console.error`, but `conflict.profileModel` (sourced from `env.ANTHROPIC_MODEL`) was interpolated raw. A crafted `ANTHROPIC_MODEL` value containing ANSI escape sequences would reach the terminal unfiltered. Fixed by wrapping it in `safeUrl()`.

Three regression tests added to `claude.plugin.conflict.test.ts` covering each vector. All 17 tests pass.

yanaSelin (approved):

yanaSelin (approved):

yanaSelin (approved):

alex-budanov:
## ⚠️ Testing correction — correct `settings.json` format

QA flagged that following the manual E2E steps above produced no warning. The root cause: the PR description says "injected a conflicting URL into `~/.claude/settings.json`" but does not show the exact format, and the natural assumption is a top-level key:

```json
{ "ANTHROPIC_BASE_URL": "https://wrong.example.com" }
```

**This format is NOT detected.** The detection code (hardened in CR-001/002/003) reads from the `env` block — which is how Claude Code itself stores env var overrides:

```json
{ "env": { "ANTHROPIC_BASE_URL": "https://wrong.example.com" } }
```

### Correct manual smoke test (Windows/PowerShell)

```powershell
# 1. Clone main and build
git clone git@gitbud.epam.com:epm-cdme/codemie-code.git codemie-code-main
cd codemie-code-main
npm install && npm run build

# 2. Back up settings.json
Copy-Item "$env:USERPROFILE\.claude\settings.json" "$env:USERPROFILE\.claude\settings.json.bak"

# 3. Inject conflicting URL under the "env" block
$s = Get-Content "$env:USERPROFILE\.claude\settings.json" | ConvertFrom-Json
if (-not $s.env) { $s | Add-Member -NotePropertyName "env" -NotePropertyValue ([pscustomobject]@{}) -Force }
$s.env | Add-Member -NotePropertyName "ANTHROPIC_BASE_URL" -NotePropertyValue "https://wrong-override.example.com" -Force
$s | ConvertTo-Json -Depth 10 | Set-Content "$env:USERPROFILE\.claude\settings.json" -Encoding UTF8

# 4. Run and observe warning on stderr
node bin\codemie-claude.js 2>&1 | Select-Object -First 30

# 5. Restore
Copy-Item "$env:USERPROFILE\.claude\settings.json.bak" "$env:USERPROFILE\.claude\settings.json" -Force
```

Expected output includes:
```
⚠️  ~/.claude/settings.json overrides detected
────────────────────────────────────────────────────────────
  Profile URL  │ <your profile url>
  Active URL   │ https://wrong-override.example.com  ← settings.json wins
```

PR #426, opened 2026-07-16
fix(analytics): correctly attribute sessions to wrapper agent names

alex-budanov:
## Summary

Fixes `codemie analytics` showing no data for `codemie-claude`, `codemie-gemini`, `codemie-codex`, `codemie-kimi`, `codemie-opencode` agents.

**Root cause:** Both session write paths stored `CODEMIE_AGENT` (short plugin name, e.g. `claude`) directly in the session JSON instead of the wrapper binary name (`codemie-claude`). Because `codemie analytics --agent codemie-claude` used an exact-match filter, wrapper sessions were invisible.

**Changes:**
- `hook.ts / createSessionRecord` — apply `toWrapperAgentName()` when writing `agentName` to the primary session file. `CODEMIE_AGENT` itself is unchanged so `AgentRegistry.getAgent()` lookups at 3 call-sites continue to work.
- `ensure-session.ts / ensureSessionFile` — second write path (used by opencode plugin) now applies the same wrapper-name transform.
- `toWrapperAgentName` — hardened: guard changed from `startsWith('codemie')` to `startsWith('codemie-')`, `codemie_xxx` underscore form normalised to `codemie-xxx`, empty-string guarded.
- `agentMatchesAnalyticsFilter` — extended; short filter `--agent claude` is broad (matches `claude` and `codemie-claude`), exact wrapper filter `--agent codemie-claude` is narrow (prevents native-loader session leakage).

## Verification

Manual end-to-end on the fix branch. Fire the hook exactly as the wrapper does:

```powershell
$sessionId = "real-test-$([DateTimeOffset]::UtcNow.ToUnixTimeSeconds())"
$payload = '{"hook_event_name":"SessionStart","session_id":"' + $sessionId + '","transcript_path":"C:/tmp/fake.jsonl","permission_mode":"default","source":"startup"}'
$env:CODEMIE_AGENT = 'claude'
$env:CODEMIE_PROVIDER = 'ai-run-sso'
$env:CODEMIE_SESSION_ID = $sessionId
$payload | node bin/codemie.js hook
```

Session file confirmed written with `"agentName": "codemie-claude"` (was `"claude"` before this fix).

Filter checks (all three verified):

| Command | Expected | Result |
|---|---|---|
| `analytics cost --agent claude --last 1d` | session found | verified |
| `analytics cost --agent codemie-claude --last 1d` | session found | verified |
| `analytics cost --agent codemie-gemini --last 1d` | not found | verified |

## Test plan

- [ ] `npx vitest run --project unit src/cli/commands/analytics/cost/__tests__/codex-agent.test.ts src/cli/commands/__tests__/hook-agent-name.test.ts src/cli/commands/analytics/__tests__/data-loader.test.ts` — 45 tests pass
- [ ] `npx vitest run --project unit` — full unit suite

### DIFF
--- a/src/agents/core/session/ensure-session.ts
+++ b/src/agents/core/session/ensure-session.ts
@@ -1,5 +1,11 @@
 import { logger } from '../../../utils/logger.js';
 
+function toWrapperAgentName(name: string): string {
+  if (!name) return name;
+  const normalized = name.toLowerCase().replace(/^codemie_/, 'codemie-');
+  return normalized.startsWith('codemie-') ? normalized : `codemie-${normalized}`;
+}
+
 /**
  * Ensure session metadata file exists for SessionSyncer.
  * Creates a new session file in ~/.codemie/sessions/ if one doesn't already exist.
@@ -23,7 +29,7 @@ export async function ensureSessionFile(
       return;
     }
 
-    const agentName = env.CODEMIE_AGENT || defaultAgentName;
+    const agentName = toWrapperAgentName(env.CODEMIE_AGENT || defaultAgentName);
     const provider = env.CODEMIE_PROVIDER || 'unknown';
     const project = env.CODEMIE_PROJECT;
     const workingDirectory = process.cwd();
--- a/src/agents/core/session/ensure-session.ts
+++ b/src/agents/core/session/ensure-session.ts
@@ -1,10 +1,5 @@
 import { logger } from '../../../utils/logger.js';
-
-function toWrapperAgentName(name: string): string {
-  if (!name) return name;
-  const normalized = name.toLowerCase().replace(/^codemie_/, 'codemie-');
-  return normalized.startsWith('codemie-') ? normalized : `codemie-${normalized}`;
-}
+import { toWrapperAgentName } from './agent-name.js';
 
 /**
  * Ensure session metadata file exists for SessionSyncer.
--- a/docs/superpowers/tasks/2026-07-16-epmcdme-10379-fix-analytics-agent-attribution/code-review.diff
+++ b/docs/superpowers/tasks/2026-07-16-epmcdme-10379-fix-analytics-agent-attribution/code-review.diff
@@ -1,0 +1,254 @@
+diff --git a/src/agents/core/session/types.ts b/src/agents/core/session/types.ts
+index 3d4f28f..b3acc5c 100644
+--- a/src/agents/core/session/types.ts
++++ b/src/agents/core/session/types.ts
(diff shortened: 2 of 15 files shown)


yanaSelin (commented):
**CR-001 — duplicate `toWrapperAgentName` in `ensure-session.ts`**

`toWrapperAgentName` is defined twice: once exported from `hook.ts` (tested via `hook-agent-name.test.ts`), and once as a private copy in `ensure-session.ts` with no test coverage. If the normalisation logic changes in one copy but not the other, the opencode session-write path will silently write wrong agent names.

Please remove the private copy and import the shared version instead. Since `ensure-session.ts` lives in the agent/session layer and importing from `src/cli/commands/hook.ts` would violate layer boundaries, the cleanest fix is to extract `toWrapperAgentName` into a small shared utility (e.g. `src/agents/core/session/agent-name.ts`) and import it in both places.

src/agents/core/session/ensure-session.ts
```diff
@@ -1,5 +1,11 @@
 import { logger } from '../../../utils/logger.js';
 
+function toWrapperAgentName(name: string): string {
+  if (!name) return name;
+  const normalized = name.toLowerCase().replace(/^codemie_/, 'codemie-');
+  return normalized.startsWith('codemie-') ? normalized : `codemie-${normalized}`;
+}
```
yanaSelin: **CR-001 — duplicate `toWrapperAgentName`**

This is a private copy of the function already exported from `hook.ts`. Two independent implementations with only one tested — if the normalisation logic changes in `hook.ts`, this path silently diverges.

Please remove this copy and import the shared version. Since importing from `src/cli/commands/hook.ts` into the agent/session layer would violate layer boundaries, the cleanest fix is to extract `toWrapperAgentName` into a shared utility (e.g. `src/agents/core/session/agent-name.ts`) and import it in both places.

alex-budanov (commented):
Done — extracted `toWrapperAgentName` into `src/agents/core/session/agent-name.ts`. Both `ensure-session.ts` and `hook.ts` now import from that shared utility. `hook.ts` re-exports it so the existing `hook-agent-name.test.ts` continues to work without changes.

src/agents/core/session/ensure-session.ts
alex-budanov: Done — extracted `toWrapperAgentName` into `src/agents/core/session/agent-name.ts`. Both `ensure-session.ts` and `hook.ts` now import from that shared utility. `hook.ts` re-exports it so the existing `hook-agent-name.test.ts` continues to work without changes.

PR #434, opened 2026-07-22
feat(proxy): add VS Code BYOK profile model support

romius:
## Summary

Add VS Code BYOK integration with the CodeMie proxy using an SSO-backed profile model.

## Motivation

Enable VS Code chat and agent workflows to use CodeMie models without exposing SSO credentials.

## Changes

- Add `codemie proxy connect vscode` with Stable and Insiders support.
- Configure VS Code language models while preserving existing settings and secrets.
- Pin proxy requests to the selected profile model and improve daemon status handling.
- Add documentation, automated coverage, and a BYOK smoke-test script.

## Configuration behavior change

This PR also adjusts the shared `ConfigLoader` profile-composition rules. Although required to make VS Code profile resolution consistent, this behavior applies to every command that uses `ConfigLoader`, not only `proxy connect vscode`.

- Profile isolation now applies whether the selected profile comes from explicit `--profile <name>` or from `activeProfile`.
- When the selected profile differs from the repository-local team profile, the local profile cannot override the selected provider, model, or credentials.
- Repository-local project context is inherited only when:
  - the selected profile does not define its own `codeMieProject` or `codeMieIntegration`; and
  - the local and selected profiles use compatible CodeMie URLs, or one URL is unspecified.
- Project and integration identifiers are never copied between explicitly different CodeMie environments.
- If the selected profile defines project context, that context remains authoritative.

This fixes cases where a globally defined active profile could previously be contaminated by a differently named local profile. Existing behavior may change for repositories that combine a global selected profile with a different local team profile.

Coverage includes active-profile selection without `--profile`, selected-profile project precedence, compatible repository overlays, cross-environment rejection, and configuration source attribution.

## Testing

Added unit and integration coverage for configuration, daemon lifecycle, and model routing.


8nevil8:
## Architecture review: prefer profile-model-as-ID over the override plugin

Thanks for the thorough docs and tests. The layering is clean (plugin registration, connector placement, error handling all follow the established patterns), but I'd like to challenge the central design before this merges.

### The concern

`ProfileModelOverridePlugin` exists solely to bridge a gap the design creates for itself: the connector writes a synthetic model ID (`codemie-profile-default`) into VS Code, and the plugin translates it back to the real model at runtime. But the plugin's only runtime input — the profile model — is **already in the connector's hands at setup time**: `connect vscode` loads `config.model` and hard-fails when it's absent. Nothing about the override needs to happen per-request.

The desktop integration proves the simpler path works: `connectors/desktop.ts` writes *real* model IDs into the client config and the proxy stays fully transparent — no rewriting machinery at all.

### Proposal

Write the profile model directly as the VS Code model `id`:

- Connector sets `id: <profileModel>` (e.g. `claude-sonnet-4-5`). It's a valid model on the same SSO gateway the profile targets — the same name CodeMie itself sends — so the proxy forwards transparently, exactly like the desktop path.
- Dedupe the managed entry by the stable `name`/`vendor` marker (`CodeMie` / `customendpoint`) instead of a fixed ID, so a profile-model change swaps the entry rather than accumulating stale ones.
- This deletes most of the PR's risk surface: `ProfileModelOverridePlugin` (+ test + registry wiring), `enforceProfileModel` on `ProxyConfig`, the `--model` / `--enforce-profile-model` daemon flags, the model/enforce fields in `DaemonState`, the `daemonMatchesRequest` complexity, and the `--use-profile-model` mode.

### The pinning argument doesn't hold up

The docs justify runtime pinning with "changing the active profile must not silently reroute a running VS Code session." With the profile model written into VS Code config, a running session keeps using the model it was configured with until the user re-runs `connect vscode` — the same guarantee, with zero runtime enforcement.

### A side effect that disappears for free

`clientType: 'vscode-byok'` currently disables four plugins that gate on hardcoded `ALLOWED_AGENTS` lists (`RequestSanitizerPlugin`, `ClaudeRequestNormalizerPlugin`, `KimiRequestNormalizerPlugin`, `CodexEncryptedContentSanitizerPlugin`) — while the managed model advertises `thinking: true` and reasoning effort, i.e. exactly the payloads those plugins handle. With no override mode, `vscode-byok` becomes purely a telemetry label and this coupling goes away.

### One assumption to verify

That VS Code's `customendpoint` vendor sends the model `id` verbatim as `body.model`. The BYOK docs say it does, and the shipped `scripts/test-vscode-byok.js` smoke test is the right tool to confirm it manually.

---

Secondary, independent of the above: the `ConfigLoader` change in `src/utils/config.ts` (basing `applyProjectOnly` on `selectedProfileName` instead of `cliOverrides?.name`, plus the new `selectedProfileDefinesProjectContext` gate) changes profile-composition semantics for **every** command, not just VS Code BYOK. It looks like a legitimate fix, but please either call it out explicitly in the PR description as a behavior change or split it into its own PR.

romius:
@8nevil8 Please review refactored version. ConfigLoader changes described in details in PR description.

package.json
```diff
@@ -42,6 +42,7 @@
     "test:all": "vitest run --project unit && vitest run --project cli && vitest run --project agent",
     "test:coverage": "vitest run --project unit --coverage",
     "test:watch": "vitest --watch",
+    "test:vscode-byok": "node scripts/test-vscode-byok.js",
```
8nevil8: should be part of test:all and/or moved to integrations/unit suite

romius: Fixed

PR #448, opened 2026-07-30
fix(agents): log !bash commands in codemie-claude session conversations (EPMCDME-13675)

SleepySML:
## Summary

Fixes [EPMCDME-13675](https://jiraeu.epam.com/browse/EPMCDME-13675) — `codemie-claude` was silently dropping `!bash` passthrough commands from the CodeMie conversation log (`~/.codemie/sessions/<sid>_conversation.jsonl`).

**Reproduction, captured first**:
- A single `!ls -al` was logged as raw XML `<bash-input> ls -al</bash-input>` instead of `!ls -al`.
- After several `!bash` commands followed by a real code question, only the FIRST bash-input made it into the log — subsequent bash commands, the follow-up question, AND the assistant's response were all silently lost.

Root cause was contained in `src/agents/plugins/claude/session/processors/claude.conversations-processor.ts`. Claude Code's `Stop` hook only fires when the assistant responds, so N bash-only turns produce zero `Stop` events. Combined with the processor's single-turn-per-invocation design and a missing `<bash-input>` unwrapper, most of the conversation was never persisted before `SessionEnd` finalized the transcript.

## Changes

- **`claude.conversations-processor.ts` — `extractCommand()`**: recognize `<bash-input>...</bash-input>` and return `!<cmd>` so User entries show what the user typed instead of leaking raw XML.
- **`claude.conversations-processor.ts` — `isSystemMessage()`**: filter `<bash-stdout>` and `<bash-stderr>` prefixes (adjacent to the existing `<local-command-stdout>` handling) so injected terminal output doesn't consume a fresh turn slot.
- **`claude.conversations-processor.ts` — `processMessages()`**: drain every pending turn per invocation in a bounded loop. `transformMessages()` is unchanged (still one turn per call); the loop wraps it so a burst of bash-only turns can no longer be truncated by the two-hook budget.
- **`config.ts` — `ConfigLoader.GLOBAL_CONFIG`/`GLOBAL_CONFIG_DIR`**: converted from class-load-time static fields to lazy getter/setter pairs with memoized overrides. Unblocks 2 pre-existing test failures in `config-project-override.test.ts` that hit any dev machine whose real `~/.codemie/codemie-cli.config.json` has `activeProfile: "default"` with empty `profiles: {}`. Production behavior unchanged; existing test workaround (direct static-write) continues to work.
- **Tests**: 6 new unit tests (`T1` single + trailing-space, `T2` stdout + stderr filter, `T3` multi-bash + question, `T4` full triple) in `claude.conversations-processor.test.ts`; integration test in `incremental-conversation-processing.test.ts` updated to assert the drain-loop's 4-record output over the existing `incremental-simple/turn-2.jsonl` fixture (old assertion of 2 records documented the bug, not correct behavior).

## Impact

Before/after over the JSONL a user's session looks like after `!ls -al`, `!pwd`, then `what files are here?`:

| Before | After |
|---|---|
| `User: <bash-input> ls -al</bash-input>` (only entry) | `User: !ls -al` |
| — | `User: !pwd` |
| — | `User: what files are here?` |
| — | `Assistant: a, b, c.` |

## Checklist

- [x] Self-reviewed via 3-lens code review (blind / edge-case / acceptance) — verdict `approve/high`, 0 blocking findings. Report: `docs/superpowers/tasks/2026-07-30-bash-input-logging/code-review-final.json`.
- [x] Manual testing: repro captured first as a failing unit test, then verified GREEN after each fix. Full spec + plan + QA report + planning artifacts committed under `docs/superpowers/tasks/2026-07-30-bash-input-logging/`.
- [x] Documentation updated: N/A — no user-facing behavior change beyond the log format.
- [x] No breaking changes.

**QA gates all PASS** (see `docs/superpowers/tasks/2026-07-30-bash-input-logging/qa-report.md`): lint, typecheck, build, unit (2400/2400), integration (196/196), commitlint. License-check locally SKIPPED due to an unrelated npm cache permission issue; CI runs it against a fresh cache.

## Acceptance criteria coverage (Jira EPMCDME-13675)

- [x] AC1 — `!` bash commands recorded in local session log
- [x] AC2 — multiple bash commands captured in correct order
- [x] AC3 — following code prompt does not lose bash entries
- [x] AC4 — log distinguishes user prompts, bash commands, assistant responses (`!` prefix + role field)
- [x] AC5 — Claude Code execution behavior unchanged
- [x] AC6 — regression tests cover `!ls -al`, `!ls -al ` (trailing space), and mixed session


src/agents/plugins/claude/session/processors/claude.conversations-processor.ts
```diff
@@ -99,54 +99,85 @@ export class ConversationsProcessor implements SessionProcessor {
...
+      let totalRecords = 0;
+      let turnsWritten = 0;
+      let lastSyncUpdate: {
+        conversations: {
+          lastSyncedMessageUuid: string;
+          lastSyncedHistoryIndex: number;
+        };
+      } | undefined;
+
+      while (true) {
```
yanaSelin: **CR-001 [major]:** Drain loop lacks explicit advance guard

**Problem:** The new `while (true)` loop in `processMessages` breaks only when `result.history.length === 0`. It does not verify that `lastProcessedMessageUuid` or `lastSyncedHistoryIndex` actually advances each iteration. If `transformMessages` ever returns a non-empty history without advancing the sync pointer (regression, malformed messages, etc.), the loop spins indefinitely inside one `processSession` call.

**Impact:** Session finalization can hang, blocking the CLI and preventing `SessionEnd` from completing cleanly.

**Recommendation:** Add an advance guard before the next iteration: break if `result.lastProcessedMessageUuid === localSync.lastSyncedMessageUuid`, or cap iterations at `session.messages.length` (or a similarly safe upper bound).

src/agents/plugins/claude/session/processors/claude.conversations-processor.ts
```diff
@@ -99,54 +99,85 @@ export class ConversationsProcessor implements SessionProcessor {
...
+          messageCount: result.history.length,
+          lastProcessedMessageUuid: result.lastProcessedMessageUuid,
+          payload: {
+            conversationId: context.agentSessionId,
+            history: result.history
+          },
+          status: CONVERSATION_SYNC_STATUS.PENDING
+        };
+
+        await appendFile(conversationsPath, JSON.stringify(payloadRecord) + '\n');
```
yanaSelin, line 174: **CR-002 [major]:** Sync pointer lags behind intermediate appends

**Problem:** Each loop iteration appends a record to disk via `appendFile` but only updates `localSync` in memory. `lastSyncUpdate` is returned and persisted by the adapter only after the full drain completes. If the process crashes or throws after some appends, the persistent sync state still points to the pre-drain position, so the next invocation re-appends already-written turns.

**Impact:** Duplicate conversation records after a crash, uncaught exception, or partial write mid-drain.

**Recommendation:** Persist the sync update per iteration so the adapter records the latest safe pointer after each append. Alternatively, refactor so a failure mid-drain does not leave records without a corresponding sync checkpoint.

src/agents/plugins/claude/session/processors/claude.conversations-processor.ts
```diff
@@ -762,7 +799,17 @@ export class ConversationsProcessor implements SessionProcessor {
...
   private extractCommand(content: string): string | null {
     const commandMatch = content.match(/<command-name>(\/[^<]+)<\/command-name>/);
-    return commandMatch ? commandMatch[1] : null;
+    if (commandMatch) {
+      return commandMatch[1];
+    }
+    // Claude Code's `!` bash passthrough is written to the transcript as
+    // `<bash-input>ls -al</bash-input>`. Unwrap to the `!<cmd>` form the user
+    // actually typed so the conversation log matches the terminal (EPMCDME-13675).
+    const bashMatch = content.match(/<bash-input>([\s\S]*?)<\/bash-input>/);
```
yanaSelin, line 864: **CR-003 [major]:** Empty or whitespace-only <bash-input> logged as bare `!`

**Problem:** `extractCommand` returns `'!' + bashMatch[1].trim()` for any `<bash-input>...</bash-input>` match, including an empty or whitespace-only captured group. This produces a User entry containing only `!`, which is not a valid command and does not represent real user input.

**Impact:** The conversation log can contain meaningless `!` entries when Claude emits an empty or whitespace-only bash-input wrapper.

**Recommendation:** Return `null` (or otherwise filter the message) when `bashMatch[1].trim()` is empty, so the message is not recorded as a bash command.

src/agents/plugins/claude/session/processors/claude.conversations-processor.ts
SleepySML: Addressed in `fea7dd3`. Replaced the unbounded `while (true)` with a `for` loop capped at `session.messages.length + 1`, and added an explicit advance guard that breaks (with a `logger.warn`) when neither `lastProcessedMessageUuid` nor `currentHistoryIndex` advances between iterations. Both fire before the next `transformMessages` call, so the loop can no longer spin indefinitely inside one `processSession`.

src/agents/plugins/claude/session/processors/claude.conversations-processor.ts
SleepySML: Addressed in `fea7dd3`. After each successful `appendFile` inside the drain loop, the processor now writes the new checkpoint (`lastSyncedMessageUuid`, `lastSyncedHistoryIndex`) back to `sessionMetadata` and calls `SessionStore.saveSession` immediately. A crash or throw after any append leaves the persistent sync state pointing at the last-written record, so the next invocation resumes past it and can't re-append. The end-of-run `applyProcessingSyncUpdates` pass still runs and remains idempotent for these fields. If the per-iteration save itself fails, we `logger.warn` and continue draining rather than aborting mid-transcript.

src/agents/plugins/claude/session/processors/claude.conversations-processor.ts
SleepySML: Addressed in `fea7dd3`. `extractCommand` now trims the captured `<bash-input>` payload and returns `null` when it's empty or whitespace-only, so we no longer emit a bare `!` User entry. As a defense in depth `isSystemMessage` also matches `/^<bash-input>\s*<\/bash-input>\s*$/` — so those messages are filtered upstream and never reach the log even as raw XML.

yanaSelin (approved):

PR #462, opened 2026-08-04
fix(assistants): fix silent zero-attachment bug in Claude session upload detection

Sergei-Nikitin-epam:
## Summary

`codemie assistants chat` was silently returning no attachments when a file was uploaded in a Claude session, because `claudeUploadsDetector` made wrong assumptions about the Claude Code JSONL message structure. Two bugs were identified via manual testing against a live Claude session (EPMCDME-13907).

**Bug A — wrong turn anchor:** The previous backward scan stopped at the first `type === 'assistant'` message. In real Claude Code JSONL the assistant `tool_use` entry appears _after_ the user's isMeta messages for the same prompt, so the scan broke before reaching the image data.

**Bug B — split isMeta structure:** Base64 image data and the `[Image: source: /path]` filename text live in _separate_ isMeta messages (not the same one), so the old single-message two-pass produced the image with a fallback filename instead of the real name.

## Changes

- **`ClaudeMessage` interface** — added `promptId?: string` (groups all JSONL entries belonging to one user prompt turn)
- **`extractFileContentFromMessages`** — replaced the backward turn-boundary scan with a promptId-based algorithm: find the most recent non-meta user message, collect all isMeta entries sharing its `promptId`, gather filenames across all of them first, then match positionally to base64 attachment items
- **Tests** — 2 new tests encoding the real split-message structure (Bug B) and the assistant-tool-use-in-between ordering (Bug A); 8 single-isMeta fixtures gained a non-meta anchor message; cross-turn isolation test given `promptId` annotations; 40 tests, all green

## Testing

- [x] Tests added for Bug A (assistant tool_use after isMeta) and Bug B (split isMeta messages)
- [x] All 40 unit tests pass (`npx vitest run`)
- [x] TypeScript clean (`npm run typecheck`)
- [x] Manual test with `node bin/codemie-claude.js` confirmed fix resolves the live session failure

## Checklist

Closes EPMCDME-13907


izhekka (approved):

8nevil8:
## Automated review — blockers

### 1. `promptId === undefined` matches all legacy sessions → false-positive stale attachments
**claudeUploadsDetector.ts — `promptId` filter**

The isMeta collection filters by `msg.promptId === anchor.promptId`, but `promptId` is typed `string | undefined`. On any JSONL file from an older Claude Code version where `promptId` is absent, every isMeta message satisfies `undefined === undefined` and matches the anchor — collecting filenames from all prior turns and producing false-positive stale attachments.

The fix for cross-turn isolation relies entirely on a field that is optional and not guaranteed by any Claude Code JSONL contract, so on real legacy sessions the bug inverts: instead of zero attachments it returns wrong attachments from previous turns — silent data corruption.

**Fix direction:** Guard explicitly for `promptId` presence before using it as the turn key. If `anchor.promptId` is `undefined`, fall back to the old assistant-message turn-boundary scan rather than matching all undefined entries.

### 2. Positional filename pairing is unguarded — deferred in the PR's own review
**claudeUploadsDetector.ts — `extractFileContentFromMessages`, positional pairing / EC-001**

Filenames are gathered in a first pass across all isMeta text items, then assigned positionally to base64 attachment content items — but no ordering contract exists between text-annotation items and attachment items within the content array.

The PR's own code review deferred this as "non-canonical ordering does not occur in practice," but there is no assertion or guard. A single Claude Code update that reorders content items silently shifts every filename by one, assigning the wrong label to every file with no error surfaced.

**Fix direction:** Match filenames to attachments by scanning the content array once and pairing each attachment item to the immediately preceding or co-located text item that matches `ATTACHMENT_PATH_PATTERN`, rather than a positional index into a separately-gathered filename list.

codemie-ai:

PR #468, opened 2026-08-06
fix(proxy): normalize Responses user identifiers

romius:
## Summary

Fixes Azure Responses API failures caused by project profiles producing user identifiers longer than Azure’s 64-character limit. The fix applies consistently to Codex, OpenCode, CodeMie Code, and VS Code BYOK traffic.

## Changes

- Generalized the VS Code-specific normalizer for all supported Responses API clients.
- Adds a short client fallback when user is missing or invalid.
- Preserves short identifiers and replaces long ones with stable 32-character SHA-256 hashes.
- Retains the existing VS Code plugin export for backward compatibility.
- Recalculates Content-Length without logging identity values.

## Impact

Before: Codex and OpenCode requests using affected project profiles could fail with Invalid 'user': string too long.

After: Responses API requests carry a bounded, stable identifier before downstream namespacing. Personal profiles and non-Responses traffic remain unchanged.

## Checklist

- [x] Self-reviewed


8nevil8:
## Automated review — policy concern

### 1. Policy blocker — user-identifier manipulation is a strictly secured operation

This PR generalizes a `user`-field normalizer (SHA-256 hash of the caller identifier) into a proxy-layer plugin that now rewrites the `user` field on all four Responses-API client types (`codemie-codex`, `codemie-opencode`, `codemie-code`, `vscode-byok`).

Any modification of user-facing identity in outbound requests — whether in headers or in the body's `user` field — is a strictly secured operation in this repo. Silently rewriting the `user` field to a SHA-256 prefix removes the ability to correlate upstream requests with the real caller identity for audit and abuse-tracing purposes, and does so for every Responses-API client at once.

**Fix direction:** Route this change through the security-review process before merging. If Azure's `user`-length limit is the underlying driver, evaluate whether the fix belongs at the client (where identity is known and a stable non-hashed identifier can be emitted) rather than at a proxy plugin that silently rewrites the field for every caller.

---

The mechanical implementation (proxy plugin placement, deterministic hex prefix, backward-compat class) is otherwise sound — the concern is the policy gate around any user-identity manipulation, not the code shape.

codemie-ai:

PR #475, opened 2026-08-10
fix(agents): stop syncing external Claude sessions on confirmed resume

maksym-govorischev:
## Summary
A confirmed external Claude session resume (`codemie-claude --resume <id>` on a session never created through CodeMie) still uploaded lifecycle metrics and could re-adopt the transcript into CodeMie's local analytics ownership index, even though conversation content sync was already suppressed.

## Motivation
EPMCDME-13367 (closed) fixed local `codemie analytics` filtering only. Anton Yeromin's comment on EPMCDME-12992 confirmed the gap: "we do show warning if resume native claude sessions, but we do not exclude them from analytics." The existing `CODEMIE_CONV_SYNC_DISABLED` flag only ever suppressed conversation content, not metrics/lifecycle events, and the ownership marker was written unconditionally on every SessionStart — silently re-adopting a transcript that had just been flagged external.

## Changes
- Added `Session.origin` (`'codemie' | 'external-resume'`), persisted on the Session record instead of relying solely on a process-local env flag.
- Added `isExternalOrigin()` predicate (checks the persisted record, falls back to env for fail-closed behavior if the record write failed).
- `AgentCLI.buildResumeEnvOverride()` now propagates `CODEMIE_SESSION_ORIGIN` instead of `CODEMIE_CONV_SYNC_DISABLED`.
- `hook.ts`: persists origin at SessionStart (including the `compact` re-entry path), skips ownership-marker writes for external-resume sessions at all call sites, and gates the two direct `MetricsSender` calls (`sendSessionStartMetrics` / `sendSessionEndMetrics`).
- `SessionSyncer.sync()`: central gate — an external-resume session skips all processors (metrics + conversations), covering every caller (hook SessionEnd, proxy timer, codex incremental sync, DesktopTelemetryRuntime) in one place.
- `native-loader.ts`: `buildOwnershipIndex()` no longer harvests `correlation.agentSessionFile` from external-resume records, so local analytics exclusion is no longer silently undone.
- `CODEMIE_CONV_SYNC_DISABLED` is left untouched — it's now a genuinely orthogonal, still-supported mechanism, unused by this flow.

## Testing
- [x] Tests added/updated (`isExternalOrigin`, `SessionSyncer.sync()` gating, `AgentCLI` resume env propagation, hook-level origin persistence + marker/metrics suppression, `native-loader` ownership-index exclusion)
- [x] `npm run typecheck` — clean
- [x] `npm run lint` — clean on touched files (repo has pre-existing unrelated `no-undef` errors in `.mjs`/`.cjs` scripts, confirmed present on `main` before this change)
- [x] `npm run build` — clean
- [x] `npm run test:unit` — 198 files, 2925 passed
- [x] `npm run test:integration` — 29 files, 204 passed
- [x] `npm run license-check` — clean
- [x] `npm run validate:secrets` — no leaks found

## Breaking Changes
None. `Session.origin` is optional and back-compat (undefined ⇒ treated as `'codemie'`) for all existing session files.

## Related Issues
Relates to EPMCDME-12992 (Jira — this repo tracks work in EPM-CDME, not GitHub Issues, so no `#NNN` applies here)

### DIFF
--- a/src/agents/core/AgentCLI.ts
+++ b/src/agents/core/AgentCLI.ts
@@ -3,6 +3,7 @@ import { readFileSync } from 'fs';
 import { join } from 'path';
 import chalk from 'chalk';
 import type { AgentAdapter, ResumeOwnershipResult } from './types.js';
+import { SESSION_ORIGIN_ENV_KEY, EXTERNAL_RESUME_ORIGIN } from './session/session-origin-audit.js';
 import { ConfigLoader, CodeMieConfigOptions } from '../../utils/config.js';
 import { ensureApiBase, DEFAULT_CODEMIE_BASE_URL } from '../../providers/core/codemie-auth-helpers.js';
 import { AuthMethod, ProviderName } from '../../providers/core/types.js';
@@ -385,12 +386,14 @@ export class AgentCLI {
               process.exit(1);
             }
 
-            // Inject into subprocess env (for lifecycle hook subprocesses that inherit it)
-            // and into the current process env (for same-process consumers such as sso syncProcessor).
+            // Inject into subprocess env — the spawned agent (and, transitively, the
+            // `codemie hook` invocations it shells out to) reads this to persist the
+            // session's origin, which is what actually gates ingestion. No same-process
+            // env write is needed here: origin is persisted to the Session record before
+            // any upload is attempted.
             Object.assign(providerEnv, buildResumeEnvOverride(true));
-            process.env.CODEMIE_CONV_SYNC_DISABLED = '1';
             appendAuditEvent('resume_external_confirmed', auditData);
-            logger.info(`[AgentCLI] External resume confirmed for agent ${this.adapter.name}; conversation sync suppressed`);
+            logger.info(`[AgentCLI] External resume confirmed for agent ${this.adapter.name}; session will not be synced`);
           }
         }
       }
@@ -406,8 +409,6 @@ export class AgentCLI {
 
       // Run the agent (welcome message will be shown inside)
       await this.adapter.run(agentArgs, providerEnv, options.printConfig ? { dryRun: true } : undefined);
-      // Clean up the process-level flag set for same-process conversation sync consumers.
-      delete process.env.CODEMIE_CONV_SYNC_DISABLED;
     } catch (error) {
       // Show user-friendly error message in console first
       const errorMessage = error instanceof Error ? error.message : String(error);
@@ -688,7 +689,7 @@ export class AgentCLI {
 }
 
 export function buildResumeEnvOverride(isExternal: boolean): Record<string, string> {
-  return isExternal ? { CODEMIE_CONV_SYNC_DISABLED: '1' } : {};
+  return isExternal ? { [SESSION_ORIGIN_ENV_KEY]: EXTERNAL_RESUME_ORIGIN } : {};
 }
 
 export function shouldBlockNonInteractiveResume(): boolean {
--- a/src/agents/core/AgentCLI.ts
+++ b/src/agents/core/AgentCLI.ts
@@ -3,7 +3,7 @@ import { readFileSync } from 'fs';
 import { join } from 'path';
 import chalk from 'chalk';
 import type { AgentAdapter, ResumeOwnershipResult } from './types.js';
-import { SESSION_ORIGIN_ENV_KEY, EXTERNAL_RESUME_ORIGIN } from './session/session-origin-audit.js';
+import { SESSION_ORIGIN, SESSION_ORIGIN_ENV_KEY } from './session/types.js';
 import { ConfigLoader, CodeMieConfigOptions } from '../../utils/config.js';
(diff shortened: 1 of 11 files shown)


src/agents/core/session/session-origin-audit.ts
```diff
@@ -3,9 +3,31 @@ import { basename, join } from 'node:path';
...
 import { logger } from '../../../utils/logger.js';
+import type { Session } from './types.js';
 
 const LOG_FILENAME = 'session-origin-audit.jsonl';
 
+/**
+ * Env var carrying the resume-ownership decision from AgentCLI into the spawned agent
+ * subprocess (and, transitively, into `codemie hook` invocations it shells out to).
+ */
+export const SESSION_ORIGIN_ENV_KEY = 'CODEMIE_SESSION_ORIGIN';
```
Dark-Sun: Let's merge SESSION_ORIGIN_ENV_KEY and EXTERNAL_RESUME_ORIGIN into one object.

So it's
SESSIONS_ORIGIN = {
CODEMIE: 'codemie-session;',
EXTERNAL_RESUME: .... 
};

src/agents/core/session/session-origin-audit.ts
```diff
@@ -3,9 +3,31 @@ import { basename, join } from 'node:path';
...
+
+/**
+ * Whether a session must never be ingested (metrics, conversations, transcript markers).
+ *
+ * Checks the persisted Session record first (authoritative — survives process boundaries
+ * like the proxy timer or DesktopTelemetryRuntime). Falls back to the env var so a session
+ * still fails closed if `createSessionRecord()` silently swallowed a write error before the
+ * origin could be persisted (it deliberately never throws — see hook.ts createSessionRecord).
+ */
+export function isExternalOrigin(session: Pick<Session, 'origin'> | null | undefined): boolean {
```
Dark-Sun, line 22: Pls explain how this function works;
1. Why don't we have one source of truth - either env or session variable?
2. :27 if (session?.origin) return false; looks redundant
3. :28 if no session.origin - return false, but if it's present - check env? Why?

src/agents/core/session/types.ts
```diff
@@ -35,6 +35,14 @@ export interface CorrelationResult {
  */
 export type SessionStatus = 'active' | 'completed' | 'recovered' | 'failed';
 
+/**
+ * Session origin — whether this session is safe to ingest.
+ * 'codemie': created through CodeMie tooling (default when unset, for back-compat with
+ * existing session files). 'external-resume': a confirmed resume of a Claude transcript
+ * that was never created through CodeMie — must never be synced (metrics or conversations).
+ */
+export type SessionOrigin = 'codemie' | 'external-resume';
```
Dark-Sun: See the other comment; Let's have these origin definition in one place

src/agents/core/AgentCLI.ts
```diff
@@ -385,12 +386,14 @@ export class AgentCLI {
...
             }
 
-            // Inject into subprocess env (for lifecycle hook subprocesses that inherit it)
-            // and into the current process env (for same-process consumers such as sso syncProcessor).
+            // Inject into subprocess env — the spawned agent (and, transitively, the
+            // `codemie hook` invocations it shells out to) reads this to persist the
+            // session's origin, which is what actually gates ingestion. No same-process
+            // env write is needed here: origin is persisted to the Session record before
+            // any upload is attempted.
             Object.assign(providerEnv, buildResumeEnvOverride(true));
```
Dark-Sun, line 394: Is overriding the env the only way?
Have you considered some config variables, maybe a singleton-like object?

Dark-Sun (changes requested):

src/agents/core/AgentCLI.ts
maksym-govorischev: We need to pass the value cross-process. There are three different processes in play Wrapper CLI -> Claude Code CLI -> Hook sub-process
The value is determined in the first one and used in the third one. 
The only other way to do this is to pre-write something on disk, but env variable option just feels better to me, because this transport approach is already used in other places in codebase.

src/agents/core/session/session-origin-audit.ts
maksym-govorischev: The logic is following:
1. If persisted session record explicitly says session is external - return true
2. If persisted session record is present and says anything else - return false, it's either codemie or something we can't parse, potentially legacy
3. If there is no session record at all, fall back to env variable. 

The last one is purely defensive mechanism as technical analysis showed that hook which writes session record can silently fail in a number of scenarios (FS write error is the simplest) and we will end up without session origin flag which by default will allow import, which is exactly what we're trying to prevent. The code is a bit convoluted, I agree, will refactor.

src/agents/core/session/session-origin-audit.ts
maksym-govorischev: Refactored

src/agents/core/session/types.ts
maksym-govorischev: Refactored

Dark-Sun:
approved

PR #482, opened 2026-08-12
feat(proxy): enhance proxy handling with noProxy rules and fallback mechanisms

crowar:
This pull request introduces several improvements to error handling, proxy configuration, and reliability in the proxy and SSO authentication code. The changes enhance resilience to upstream failures, improve diagnostics and recovery for authentication issues, and add robust support for proxy bypass rules (NO_PROXY) including reading from `.npmrc`. The most important changes are:

**Reliability and Fallbacks:**
* The `fetchClaudeModels` function now falls back to a preferred list of Claude model IDs if the proxy upstream returns a 5xx error, improving reliability when the model catalog is unavailable. [[1]](diffhunk://#diff-9501df30b1ebc6835738899f2c340e827b6a61aa807cede2c8cc54cde055e148R96-R116) [[2]](diffhunk://#diff-42f9ba044d326a13e7995c9001ede7983c3366bd9fb37189d7ebca9e55ec9c54R236-R248)

**Proxy and Network Configuration:**
* The `ProxyHTTPClient` class now fully supports NO_PROXY rules, including parsing environment variables and `.npmrc` for host/domain/CIDR rules, and routes requests directly when a match is found. This ensures correct proxy bypass behavior for both HTTP and HTTPS, and logs rule loading for diagnostics. [[1]](diffhunk://#diff-d4a4241f2b1d1ca41dacb532fe2f10402c69d8ed461316ad1f2173c720304186R12-R15) [[2]](diffhunk://#diff-d4a4241f2b1d1ca41dacb532fe2f10402c69d8ed461316ad1f2173c720304186R50-R245) [[3]](diffhunk://#diff-d4a4241f2b1d1ca41dacb532fe2f10402c69d8ed461316ad1f2173c720304186L96-R257) [[4]](diffhunk://#diff-d4a4241f2b1d1ca41dacb532fe2f10402c69d8ed461316ad1f2173c720304186L280-R445)
* Refactored agent selection in `ProxyHTTPClient` to distinguish between direct and proxy agents, and to properly clean up all agents on close. [[1]](diffhunk://#diff-d4a4241f2b1d1ca41dacb532fe2f10402c69d8ed461316ad1f2173c720304186R50-R245) [[2]](diffhunk://#diff-d4a4241f2b1d1ca41dacb532fe2f10402c69d8ed461316ad1f2173c720304186L280-R445)

**Authentication and Error Handling:**
* The SSO authentication plugin now validates cookie header values for invalid characters, skips invalid entries, and throws a clear error if all cookies are invalid, guiding the user to re-authenticate.
* Improved normalization of errors: malformed header/cookie errors are now mapped to an `AuthenticationError` with a recovery message, and TLS certificate errors are mapped to a specific `UpstreamError` for better user guidance.

**State Management:**
* When the proxy health check transitions from unhealthy to healthy, any stale health status is cleared from persisted state, ensuring accurate status reporting.

These changes collectively improve the robustness, debuggability, and user experience of the proxy and SSO authentication system.


8nevil8:
## Automated review — critical findings

### 1. `NO_PROXY` entries with a port are silently ignored
**proxy-http-client.ts — `parseNoProxyRules` / `shouldBypassProxy`**

NO_PROXY entries that include a port (e.g. `example.com:8080`, `10.0.0.1:443`) are stored verbatim as `{ kind: 'host', value: 'example.com:8080' }`, but `getAgentForUrl` passes `url.hostname` — which never contains a port — so these rules silently never fire and the proxy is never bypassed for them.

Standard NO_PROXY semantics for `host:port` entries ("bypass only for this port") are completely unimplemented. Every such rule is a silent no-op. Requests users expect to go direct are still routed through the proxy.

**Fix direction:** Strip the port suffix from rule values during parsing to store `{ kind: 'host', value: 'example.com', port: 8080 }`, then match against `url.hostname` + `url.port` in `shouldBypassProxy`.

### 2. `fetchClaudeModels` silently swallows 5xx errors and returns stale hardcoded list
**desktop.ts: `fetchClaudeModels` ~line 108**

On any 5xx response from the upstream model catalog, the function returns the hardcoded `PREFERRED_CLAUDE_MODELS` array to the caller with no thrown error and no way for the caller to know the list is stale.

A degraded or mis-configured upstream causes Claude Desktop to silently present a frozen model list that drifts as models are released or deprecated. The caller has no signal to distinguish live from fallback, so errors stay hidden until a user notices wrong/missing models.

**Fix direction:** Either throw and let the caller decide the fallback policy, or return a discriminated result (e.g. `{ models, source: 'fallback' | 'live' }`) so the caller can surface a warning to the user.

8nevil8 (approved):

PR #483, opened 2026-08-13
fix(cli): normalize root base-url to code-assistant-api path

AntonYeromin:
## Summary
`--base-url` (or a stored SSO/JWT profile's `codeMieUrl`) only worked when pointed at the full `/code-assistant-api` path. Passing the root CodeMie host (e.g. `https://codemie.lab.epam.com`) caused `API Error: 405 Not Allowed`, because requests went to `/v1/messages` on the root domain instead of `/code-assistant-api/v1/messages`.

## Changes
- `src/agents/core/AgentCLI.ts`: normalize `config.baseUrl` with `ensureApiBase()` whenever `config.provider` is `ai-run-sso` or `bearer-auth` (the two CodeMie-native auth routes that proxy through the local CodeMie proxy). Idempotent — an already-correct `/code-assistant-api` URL is left untouched. Keyed on `provider` rather than `authMethod` because stored SSO profiles never populate `authMethod`.
- `docs/AUTHENTICATION.md`: document that root or `/code-assistant-api` URLs are both accepted, and add a troubleshooting entry for the `405 Not Allowed` error.

## Testing
- [x] Manual testing done — verified end-to-end against a live CodeMie SSO backend (`codemie.lab.epam.com`) with `CODEMIE_DEBUG=true`, confirming `/v1/messages` now resolves through `/code-assistant-api` and returns `200 OK` instead of `405`.
- [ ] Tests added/updated (none requested)

## Checklist

Refs: [EPMCDME-14096](https://jiraeu.epam.com/browse/EPMCDME-14096)

### DIFF
--- a/src/agents/core/AgentCLI.ts
+++ b/src/agents/core/AgentCLI.ts
@@ -224,6 +224,18 @@ export class AgentCLI {
         config.authMethod = AuthMethod.JWT;
       }
 
+      // Normalize a user-supplied --base-url (or one loaded from a profile) so the
+      // required /code-assistant-api path is always present, even when the root
+      // CodeMie URL is passed. Idempotent: leaves an already-correct URL untouched.
+      // Keyed on provider (not authMethod) because stored SSO profiles never set
+      // authMethod — proxy routing there is decided by provider.authType === 'sso'.
+      if (
+        config.baseUrl
+        && (config.provider === ProviderName.AI_RUN_SSO || config.provider === ProviderName.BEARER_AUTH)
+      ) {
+        config.baseUrl = ensureApiBase(config.baseUrl);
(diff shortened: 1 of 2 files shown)


m-golovchin (approved):

src/agents/core/AgentCLI.ts
```diff
@@ -224,6 +224,18 @@ export class AgentCLI {
         config.authMethod = AuthMethod.JWT;
       }
 
+      // Normalize a user-supplied --base-url (or one loaded from a profile) so the
+      // required /code-assistant-api path is always present, even when the root
```
8nevil8, line 228: it will break client deployments if they deploy under different subpath

PR #484, opened 2026-08-13
fix(agents): codemie-claude hook command-not-found and unsupported effort param (EPMCDME-14035)

SleepySML:
## Summary

Fixes two independent `codemie-claude` failures reported in [EPMCDME-14035](https://jiraeu.epam.com/browse/EPMCDME-14035):

1. Claude Code hooks failing with `/usr/bin/bash: line 1: codemie: command not found`.
2. Requests to `claude-4-5-sonnet` failing with `API Error: 400 "This model does not support the effort parameter"`.

Both were reproduced before fixing.

## Root causes

- **Bug 1 (hooks):** Installed hook configs invoked the **bare `codemie`** command. Claude Code runs hooks in a fresh `bash -c "codemie hook"` shell; when that shell's PATH doesn't include the codemie bin dir (a user-prefix install without admin rights), every hook fails with command-not-found.
- **Bug 2 (effort):** The SSO proxy `claude-request-normalizer` only ever **added** `output_config.effort` for adaptive-thinking models and **never stripped** it for models that don't support it. Newer Claude Code emits `effort` from its `--effort` flag, so a `claude-4-5-sonnet` request carried `effort` through untouched → upstream 400.

## Changes

**Bug 1 — absolute, PATH-independent hook commands**
- `src/utils/hook-command.ts` *(new)* — shared resolver: `resolveCodemieBinary()` (`getCommandPath('codemie')` → `process.argv[1]` fallback; quotes paths with spaces; prefixes `node` for a Windows `.js` fallback), `resolveHookCommand()` (rewrites the leading `codemie` token), `rewriteHooksCommandTree()` (walks a hooks config). Resolution never throws.
- `src/agents/core/extension/BaseExtensionInstaller.ts` — post-copy `localizeInstalledHooks()` rewrites the installed `hooks.json` to an absolute path (covers **Claude and Gemini** via the shared base class). Non-fatal.
- `src/agents/plugins/codemie-code.plugin.ts` — inline `OPENCODE_HOOKS` default hooks (`buildDefaultHooks`) use the resolved absolute path.
- `src/migrations/006-resolve-hook-command-paths.migration.ts` *(new)* — one-time repair of already-installed Claude/Gemini hooks for users whose plugin version didn't bump. Idempotent; returns failure on a write error so the runner retries.

**Bug 2 — strip unsupported `effort`**
- `src/providers/plugins/sso/proxy/plugins/claude-request-normalizer.plugin.ts` — new `handleUnsupportedEffort()` strips `output_config.effort` and any top-level `effort` for models **not** matching the adaptive-thinking patterns; runs regardless of whether `thinking` is present. Adaptive models (opus-4-7+, sonnet-5) keep `effort`.

## Testing

- **20 new unit tests**; full unit suite **3037 passing**.
- `lint`, `typecheck`, `build`, `commitlint`, and the **gitleaks** secrets scan all green.
- `license-check` and the integration/agent suites were skipped **locally** for environmental reasons (npm-cache perms / live-SSO / env-heavy) — all enforced by CI.

## Code review

A three-lens review (blind / edge-case / acceptance) ran on the diff and this PR **fixes** what it found:
- 🔴 **critical** — migration marked itself permanently applied on a write failure (never retried) → now returns failure so it retries next launch.
- 🟠 **major** — Windows `.js` `argv[1]` fallback wasn't directly invocable as a hook command → now `node`-prefixed.

## Acceptance criteria

- [x] `codemie` available to hook execution after `codemie-claude` setup
- [x] `SessionStart` hooks no longer fail with command-not-found
- [x] `UserPromptSubmit` hooks no longer fail with command-not-found
- [x] `claude-4-5-sonnet` requests no longer include unsupported `effort`
- [x] Unsupported params omitted safely before send
- [x] Works without admin rights to downgrade CodeMie

## Notes

- Squash-and-Merge per project default.
- Hook rewriting and migration 006 are idempotent and non-fatal — they never break install or startup.

### DIFF
--- a/docs/superpowers/tasks/2026-08-12-codemie-claude-hook-path-and-effort-param/actual-complexity.json
+++ b/docs/superpowers/tasks/2026-08-12-codemie-claude-hook-path-and-effort-param/actual-complexity.json
@@ -1,0 +1,97 @@
+{
+  "task": "EPMCDME-14035 bugfix: shared hook-command path resolver (src/utils/hook-command.ts) wired into BaseExtensionInstaller and codemie-code.plugin with one-time migration 006 for existing users, plus handleUnsupportedEffort handler added to SSO proxy claude-request-normalizer to strip the effort parameter for non-adaptive Claude models",
+  "generated": "2026-08-13T00:00:00.000Z",
+  "dimensions": {
+    "component_scope": {
+      "score": 5,
+      "label": "XL",
+      "affected": "hook-command utility, BaseExtensionInstaller base class, codemie-code.plugin, migration 006 + index, claude-request-normalizer proxy plugin",
+      "layers": "Service, Agent-Tool, DB-Persistence, External"
+    },
+    "requirements_clarity": {
+      "score": 2,
+      "label": "S",
+      "status": "Clear",
+      "gaps": null
+    },
+    "technical_risk": {
+      "score": 3,
+      "label": "M",
+      "risk_factors": "migration 006 rewrites existing user hook JSON files (data migration via migration runner); hook path resolution must never throw (handled with graceful fallback chain)",
+      "mitigation": "localizeInstalledHooks is non-fatal (catches all errors, logs warning); migration returns success:false on write failure so runner retries; resolveCodemieBinary falls back through three levels before returning bare 'codemie'"
+    },
+    "file_change_estimate": {
+      "score": 5,
+      "label": "XL",
+      "modified_files": 4,
+      "modified_file_list": [
+        "src/agents/core/extension/BaseExtensionInstaller.ts",
+        "src/agents/plugins/codemie-code.plugin.ts",
+        "src/migrations/index.ts",
+        "src/providers/plugins/sso/proxy/plugins/claude-request-normalizer.plugin.ts"
+      ],
+      "new_files": 7,
+      "new_file_list": [
+        "src/utils/hook-command.ts",
+        "src/utils/__tests__/hook-command.test.ts",
+        "src/agents/core/extension/__tests__/BaseExtensionInstaller.hooks.test.ts",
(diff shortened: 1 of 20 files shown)


8nevil8:
## Automated review — critical finding

### 1. Migration hardcodes `~/.codemie` instead of using `getCodemiePath()`
**src/migrations/006-resolve-hook-command-paths.migration.ts:34**

`hookFiles()` hard-codes `path.join(homedir(), '.codemie', ...)` instead of calling `getCodemiePath()` from `src/utils/paths.ts`.

`AGENTS.md` Common Pitfalls explicitly forbids this. Users with a custom codemie home directory (via env var or non-default install prefix) will have the migration silently skip their hook files — leaving the command-not-found bug unrepaired for exactly the users most likely affected (non-admin, user-prefix installs).

**Fix direction:** Replace `path.join(homedir(), '.codemie', 'claude-plugin', 'hooks', 'hooks.json')` with `path.join(getCodemiePath(), 'claude-plugin', 'hooks', 'hooks.json')` to honour the project's path abstraction.

---

Both fixes are otherwise architecturally sound and correctly layered — the hook-path resolver lands in `src/utils/`, migration 006 follows the established runner pattern, and effort stripping lands in the correct proxy-plugin chain. Only the hardcoded path violates a documented project guardrail.

SleepySML:
Done — addressed in `9d92a73`. `hookFiles()` now uses `getCodemiePath('claude-plugin', 'hooks', 'hooks.json')` (which honours `CODEMIE_HOME`) instead of the hardcoded `homedir()/.codemie`, so custom/user-prefix codemie homes are repaired too. The `.gemini` extensions path is left on `homedir()` since it targets a different tool's home that `getCodemiePath()` does not cover. Migration 006 unit tests still green (4/4).

PR #486, opened 2026-08-17
fix(analytics): exclude external-resume sessions from analytics reports

maksym-govorischev:
## Summary
- Follow-up to #475: a confirmed external-resume session's own metrics upload was correctly gated, but the analytics report pipeline (`MetricsDataLoader` → cost enrichment → aggregator) had no awareness of session `origin` and silently resurrected such sessions into reports.
- Also fixes the confirmation prompt's copy, which still claimed metrics "WILL be tracked" — stale wording from before lifecycle metrics were gated for this case.

## Changes
- `src/cli/commands/analytics/data-loader.ts`: `MetricsDataLoader.loadSession()` now drops any session with `origin === 'external-resume'` before it reaches cost enrichment or aggregation, so it can never be rescued back into a report via the zero-delta/real-cost `keepSessionIds` exception.
- `src/agents/core/AgentCLI.ts`: reworded `promptExternalResume()`'s warning to state plainly that the session will not be tracked as a CodeMie session (no metrics, no transcript sync, not in Analytics), while noting API calls still route through the CodeMie proxy. Dropped the redundant fallback-command line.

## Testing
- [x] `npm run typecheck`
- [x] `npx vitest run` on `data-loader.test.ts` and `AgentCLI-resume.test.ts` (existing suites, all passing)
- [x] Manual testing done

## Checklist

Refs EPMCDME-12992

### DIFF
--- a/src/agents/core/AgentCLI.ts
+++ b/src/agents/core/AgentCLI.ts
@@ -671,7 +671,11 @@ export class AgentCLI {
     console.log(chalk.white('If you continue:'));
     console.log(chalk.white('  • This session will NOT be tracked as a CodeMie session — no CodeMie session metrics, no transcript sync, and it will not appear in CodeMie Analytics.'));
     console.log(chalk.white(`  • API calls still route through the CodeMie proxy while using ${wrapperCommand}, so usage will still be logged/billed there.\n`));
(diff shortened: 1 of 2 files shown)


Dark-Sun:
approved

PR #488, opened 2026-08-17
feat(analytics): add auth status gate for metrics sync

8nevil8:
## Summary
Expired CodeMie SSO cookies caused the metrics endpoint (fronted by Keycloak) to answer POSTs with HTTP 200 + HTML login page instead of 401. Analytics auth failures were completely silent — sessions vanished from the metrics dashboard while local JSONL deltas looked healthy. This adds a three-layer auth status gate so broken analytics auth is detected and surfaced immediately.

## Changes
- `src/utils/analytics-auth-status.ts` (new): file-based marker recording metrics-endpoint credential rejections (`~/.codemie/analytics-auth-status.json`); best-effort, never throws
- `metrics-api-client.ts`: classify HTTP 401/403 and non-JSON HTML responses (Keycloak login page) as auth failures; write marker on rejection, clear on successful send
- `sso.auth.ts`: clear the marker on successful `codemie profile login`
- `hook.ts`: generic `UserPromptSubmit` gate (all agents via unified `codemie hook`) — blocks the prompt with `codemie profile login` instructions when analytics auth is missing or was server-rejected; active only when analytics sync is configured
- `AgentCLI.ts`: launch-time analytics auth check for non-SSO providers when analytics is enabled (`codeMieUrl` set); reuses SSO `validateAuth` (stored creds + live API test) with interactive re-auth prompt; non-fatal on decline; clears stale marker on success

## Impact
Before: expired analytics credentials → silent metrics loss for days (observed 08-12 → 08-14 locally; data recovered only after manual re-login + session resumes).
After: launch-time warning with re-auth prompt; hard block at prompt level with exact re-login command if still unresolved; marker cleared automatically by login or any successful send.

## Testing
- [ ] Tests added/updated
- [x] Manual testing done (typecheck, ESLint zero-warning, build, lint-staged + related vitest via pre-commit)

## Checklist

### DIFF
--- a/.ai-run/guides/security/security-practices.md
+++ b/.ai-run/guides/security/security-practices.md
@@ -15,6 +15,69 @@ Security patterns for CodeMie Code: credential storage, data sanitization, input
 | Data sanitization | Auto-redact sensitive data in logs |
 | Secure storage | Encrypted credential storage (keytar + fallback) |
 | Least privilege | File permissions, scoped access |
+| **Request header integrity (project & user attribution)** | **Any add/remove/rewrite of project- or user-attribution headers is a CRITICAL change — security review required before merge** |
+
+---
+
+## CRITICAL: Project & User Attribution Headers
+
+Any change that adds, removes, renames, or rewrites a header carrying **project identity**, **user identity**, or **session attribution** on outbound requests is a **CRITICAL, security-reviewed change**. This applies to headers, to fields that stand in for them (e.g. the Responses API `user` body field, JWT `sub` overrides), and to any derivation source that feeds them.
+
+These identifiers back audit trails, tenant isolation, per-user budget accounting, abuse tracing, and analytics attribution. A silent rewrite — even a "harmless" hash, prefix, or normalization — can:
+
+- break audit correlation between CLI action and server-side event,
+- misattribute usage or billing across tenants/users,
+- hide abuse behind an unreachable identifier,
+- cross-attribute analytics between repositories/branches.
+
+### Headers and identifier fields in scope
+
+Injected today by `src/providers/plugins/sso/proxy/plugins/header-injection.plugin.ts`:
+
+| Header | Meaning |
+|---|---|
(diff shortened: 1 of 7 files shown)


8nevil8:
## Automated review — critical findings

### 1. `ssoSetupSteps.validateAuth(config)` — live network call on every agent launch
**src/agents/core/AgentCLI.ts:295–310**

Any slow response or unreachable CodeMie backend hangs agent startup with no visible timeout guard. Every `anthropic-subscription` / `moonshot-subscription` user takes the hit on every launch.

**Fix direction:** Skip the live API call here; rely solely on the marker written by `MetricsApiClient` (already present), or gate on the marker first and only call `validateAuth` when the marker already exists.

### 2. Programmatic mode re-throws all gate errors, not just intentional blocks
**src/cli/commands/hook.ts:550–570**

In programmatic mode (`config` defined), the outer `catch` unconditionally re-throws all gate errors. A transient filesystem error reading the credential store or the auth-status file surfaces to the VSCode host as a blocking error, violating the gate's own "auxiliary, must never break the prompt flow" contract.

**Fix direction:** Distinguish intentional blocks (`isAuthBlock` flag or sentinel error subclass) from unexpected exceptions; only re-throw intentional blocks in programmatic mode.

### 3. Gate blocks on missing credentials before any server rejection
**src/cli/commands/hook.ts:523–543**

The gate blocks when `hasValidAuth = false` AND `authStatus = null` — i.e. it blocks on **missing credentials** even when no server-side rejection has ever been recorded. A fresh install with `CODEMIE_URL + CODEMIE_SYNC_API_URL` in the environment but no prior login gets hard-blocked at every prompt before the first metrics send attempt. This contradicts the gate's stated contract ("blocks when auth is **known broken**").

**Fix direction:** Restrict the credential-absence block to cases where a marker exists, or only run the credential check when `authStatus` is already set; let the first send attempt write the marker if credentials are absent.

TarasSpashchenko (approved):

PR #507, opened 2026-08-26
feat(proxy): enhance VS Code model handling with gateway integration and fallback logic

crowar:
This pull request enhances how the VS Code connector discovers and configures available language models, making the process more robust and informative. The main improvement is that the connector now queries the local gateway for the actual set of available models, falls back gracefully to a curated catalog if discovery fails, and always returns the list of configured model IDs. Logging and error handling have also been improved to aid debugging and transparency.


romius:
@crowar 
Thanks for the work - dynamic model discovery is the right direction. However, I think we should hold this PR for now.
This implementation retrieves only model IDs and derives the remaining attributes by matching the existing local catalog. For new models, it falls back to generic chat-completions defaults, which can configure Messages or Responses models incorrectly.
Full model metadata required by VS Code-API type, capabilities, token limits, reasoning, vision, tool calling, etc. - is being implemented on the gateway backend already. Once that contract is available, VS Code should consume the backend model definitions directly and switch fully to backend-driven configuration.
The change should be ready soon, so I suggest revisiting this PR after the backend changes are available rather than merging the current ID-only mapping approach.

codemie-ai:
[AUTO_CLOSE_WARNING] ⏰ This pull request is older than 7 days and will be automatically closed in 7 more days (when it reaches 14 days old)! To maintain this PR, either convert it to Draft or complete your changes and merge.

codemie-ai:
[AUTO_CLOSE_NOTICE] 🔒 Closed automatically: no activity for 15 days (last activity: 2026-08-28).

Nothing is lost — the branch is untouched. Reopen the PR to continue; housekeeping will give it a fresh 7/14-day cycle starting from the reopen moment.

PR #511, opened 2026-08-27
feat(fcc): add Free Claude Code provider for custom LLM connections

00062541:
## Summary

Adds an opportunity to use Claude CLI - like UI in the Command Pallete with models that can be different from the Anthropics

## Changes

New CLI
## Checklist

- [x] Self-reviewed

### DIFF
--- a/docs/FCC-SETUP.md
+++ b/docs/FCC-SETUP.md
@@ -1,0 +1,365 @@
+# FCC (LiteLLM Gateway) Setup Guide
+
+## For Custom LLM Deployments
+
+This guide explains how to set up and use FCC (Free Claude Code), a custom Claude Code deployment that routes requests through a LiteLLM gateway with custom authentication.
+
+---
+
+## Table of Contents
+
+1. [Prerequisites](#1-prerequisites)
+2. [Quick Start](#2-quick-start)
+3. [Manual Installation](#3-manual-installation)
+4. [Configuration](#4-configuration)
+5. [Usage](#5-usage)
+6. [Troubleshooting](#6-troubleshooting)
+
+---
+
+## 1. Prerequisites
+
+Before installing FCC, ensure you have:
+
+### Required Access
+
+- **LiteLLM Gateway URL**: The URL of your LiteLLM proxy server
+- **API Key**: Your LiteLLM API key (obtain from your infrastructure team)
+- **Auth Token**: Authentication token if required by your deployment
+
+### System Requirements
+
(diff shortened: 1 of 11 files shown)


8nevil8:
Thanks for the work here — but after review I'm closing this in favor of the existing **`litellm`** provider, because the `fcc` provider duplicates it rather than adding new capability.

### Why it's a duplicate

FCC and `litellm` target the exact same mechanism. Claude Code routes to a LiteLLM gateway via two env vars — `ANTHROPIC_BASE_URL` (the proxy) and `ANTHROPIC_AUTH_TOKEN` (the bearer key) — and both providers do only that. The PR's own doc is even titled *"FCC (LiteLLM Gateway) Setup Guide."* The plugin also copies the `litellm` plugin layout 1:1 (`*.template.ts`, `*.setup-steps.ts`, `*.models.ts`, `index.ts`) and hits the same `/v1/models` endpoint.

Where `fcc` differs from `litellm`, it's unfortunately weaker:

| | `litellm` (existing) | `fcc` (this PR) |
|---|---|---|
| Credentials | 1 key (matches how LiteLLM auth works) | 2 prompts, but `fccLiteLLMKey` is wired to nothing |
| Models | Dynamic `/v1/models` discovery | Hardcoded list with invalid IDs (e.g. `claude-opus-4-5-20250929`) |
| `authType` | `api-key` (accurate) | `'sso'` — but there is no SSO flow |
| Claude compatibility | Listed in `claude.supportedProviders` | Not listed → `codemie-claude --provider fcc` is runtime-rejected |
| Tests | Yes | None |

There's also an unrelated change bundled in — the `detectLiteLLMEnforcement()` rewrite removes the "leave blank to skip" option, forcing every `ai-run-sso`/`litellm` setup through a mandatory SSO round-trip. That's a separate concern and shouldn't ride along with a new provider.

### Recommendation

Use the existing **`litellm`** provider for custom LLM / gateway connections:

```bash
codemie setup            # choose "LiteLLM" as the provider
# Proxy URL:  https://your-litellm-gateway.example.com
# API Key:    <your gateway key>
```

The genuinely useful parts of this PR — the corporate-proxy troubleshooting guide and the winget prerequisite installer (uv/Node/Claude CLI) — don't require a new provider and could be contributed on top of `litellm` in a focused follow-up. If we need FCC branding, a thin display-name alias over `litellm` is preferable to a second implementation that will drift.

Closing as duplicate. Happy to help repackage the installer/docs piece if useful.

PR #528, opened 2026-09-03
fix(analytics): fix Gemini agent analytics tracking (EPMCDME-14674)

BogLesh:
## Ticket
[EPMCDME-14674](https://jiraeu.epam.com/browse/EPMCDME-14674)

## Summary
When running codemie analytics --report --last 8d, users utilizing Gemini agents received "No analytics data available". This PR resolves parsing, deduplication, and attribution gaps for Gemini CLI transcripts:

- **Streaming JSONL Parsing (GeminiSessionAdapter):** Added .jsonl discovery, header extraction, and line-by-line parsing for streaming chat transcripts (session-*.jsonl), handling $set.messages and raw turn chunks without throwing SyntaxError.
- **In-Place Turn Deduplication:** Deduplicated turn-level chunks by message id in parseSessionFile, merging partial streaming updates (content, tokens, thoughts, toolCalls) to prevent token double-counting.
- **Defensive Prompt Text Extraction (GeminiMetricsProcessor):** Added extractPromptText() supporting strings, structured arrays (content: [{ text: "..." }]), and stripping internal <session_context> XML blocks.
- **Git Branch Attribution:** Bound active gitBranch to generated MetricDelta records.
- **Serena MCP Tool Mapping:** Mapped Serena MCP tools (mcp_serena_*) to standard file operations (write, edit, 
ead, delete) and extracted line counts across varied argument schemas.
- **Model Pricing Table (pricing.json):** Added standard dashed entries (gemini-3-7-flash, gemini-3-5-flash, gemini) and dotted aliases to support lookupPrice() normalization.

## Verification
- Vitest unit tests: **80 passed / 80 tests (100%)** (
pm run test:unit)
- Typecheck & Lint: 	sc --noEmit and eslint passed with 0 errors / 0 warnings
- Build verification: 
pm run build compiled successfully
- Real-world validation: codemie analytics --report --last 8d --report-format json successfully parsed real Gemini CLI sessions, attributing models, costs, and Serena MCP tool calls.

-----------------
### Note on MCP Tool Mapping & Proposed Follow-Up (Technical Debt)

In this PR, I added explicit mappings for Serena MCP tools (`mcp_serena_*`) because Serena was the specific MCP server utilized in my Gemini CLI workflows where the missing file operation metrics were identified.

**Known Limitation & Technical Debt:**
Hardcoding specific MCP server names (`mcp_serena_*`) solves the immediate issue for this stack, but it introduces technical debt: developers or external contributors using alternative MCP servers (such as `@modelcontextprotocol/server-filesystem` or custom IDE MCP extensions) will have their tool calls recorded in general usage tables, but their file edits and line diffs will be omitted unless explicitly mapped.

**Proposed Solution (Universal / MCP-Agnostic Extraction):**
I propose creating a follow-up task/PR to make file operation extraction truly generic and MCP-agnostic by:

1. **Stripping dynamic MCP prefixes:** Normalizing incoming tool names (e.g., `toolName.replace(/^mcp_[a-z0-9_-]+_/i, '')`), so that `mcp_serena_write_file`, `mcp_filesystem_write_file`, and native `write_file` all resolve to a common canonical name.
2. **Semantic intent matching:** Classifying operations via action patterns (`/write|create/`, `/replace|edit|patch/`, `/read|view/`, `/delete/`) and resolving file paths/diffs across standard argument conventions (`path`, `filePath`, `relative_path`, `content`, `replacement`, `needle/repl`).

This will ensure that any current or future MCP coding server works out of the box with zero manual config required.

### DIFF
--- a/src/agents/plugins/gemini/session/processors/gemini.metrics-processor.ts
+++ b/src/agents/plugins/gemini/session/processors/gemini.metrics-processor.ts
@@ -17,6 +17,37 @@ import type { MetricDelta } from '../../../../core/metrics/types.js';
 import { extractFormat, detectLanguage } from '../../../../../utils/file-operations.js';
 
 /**
+ * Strip <session_context>...</session_context> blocks and trim the result.
+ */
+function filterSessionContext(text: string): string | null {
+  const cleaned = text.replace(/<session_context>[\s\S]*?<\/session_context>/gi, '').trim();
+  return cleaned || null;
+}
+
+/**
+ * Extract user prompt text from varied content formats.
+ */
+function extractPromptText(content: unknown): string | null {
+  if (typeof content === 'string') {
+    return filterSessionContext(content);
+  }
+  if (Array.isArray(content)) {
+    const textParts = content
+      .map((part) => {
+        if (part && typeof part === 'object' && typeof part.text === 'string') {
+          return part.text;
+        }
+        return '';
+      })
+      .filter(Boolean);
+    if (textParts.length > 0) {
+      return filterSessionContext(textParts.join('\n'));
+    }
+  }
+  return null;
+}
+
+/**
  * Gemini message structure (from gemini.session-adapter.ts)
  */
 interface GeminiMessage {
@@ -74,7 +105,7 @@ export class GeminiMetricsProcessor implements SessionProcessor {
     try {
       logger.info(`[${this.name}] Transforming ${session.messages.length} messages to deltas`);
 
-      const deltas = this.transformMessagesToDeltas(session);
+      const deltas = this.transformMessagesToDeltas(session, _context);
 
       if (deltas.length === 0) {
         logger.debug(`[${this.name}] No deltas generated from messages`);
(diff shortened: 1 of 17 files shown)


src/utils/pricing.json
```diff
@@ -329,6 +329,36 @@
...
+    "cacheRead": 0.05,
+    "cacheWrite": 0.5
+  },
+  "gemini-3.5-flash": {
+    "input": 0.5,
+    "output": 3.0,
+    "cacheRead": 0.05,
+    "cacheWrite": 0.5
+  },
+  "gemini": {
```
yanaSelin: This bare, version-less key becomes a silent fallback for *any* unrecognized `gemini-*` model, not just a placeholder. `lookupPrice()` (`src/utils/pricing.ts`) does exact match first, then falls back to the **longest key that segment-matches** the normalized model name (`isSegmentMatch`). Since `"gemini"` segment-matches any `gemini-<anything>` name (e.g. a future `gemini-4-ultra` with no dedicated entry), it will resolve to this catch-all and get priced as Flash tier (0.5/3.0) instead of returning `null`/unpriced. Every other family in this table (`grok-*`, `o1*`, `claude-*`) avoids a bare tier key for exactly this reason — Claude even has a dedicated `claudeTierFallback()` that logs "inexact" rather than silently matching a short key. Suggest dropping this entry, or adding an explicit gemini tier-fallback (logged as inexact) instead of relying on `isSegmentMatch` here.

src/agents/plugins/gemini/session/processors/gemini.metrics-processor.ts
```diff
@@ -108,28 +139,32 @@ export class GeminiMetricsProcessor implements SessionProcessor {
...
       if (msg.type === 'user') {
-        lastUserPrompt = msg.content;
+        lastUserPrompt = extractPromptText(msg.content);
         continue; // User messages don't generate deltas
       }
 
-      // Process assistant (gemini) messages
-      if (msg.type === 'gemini') {
+      // Process assistant (gemini or assistant) messages
+      if (msg.type === 'gemini' || (msg.type as string) === 'assistant') {
```
yanaSelin, line 161: `transformMessagesToDeltas` now treats `msg.type === 'assistant'` the same as `'gemini'` when generating cost/token deltas: `if (msg.type === 'gemini' || (msg.type as string) === 'assistant') {`. But `GeminiConversationsProcessor`'s turn detection (`turn-detector.ts:66`, `geminiMessages: turnMessages.filter(m => m.type === 'gemini')`) still only recognizes `'gemini'`. If real Gemini JSONL sessions actually emit `type: 'assistant'` messages (which is presumably why this branch was added), analytics will report cost/tokens for turns whose content never makes it into the synced conversation history — a discrepancy between the two processors. Please either align `turn-detector.ts`'s type set with this change, or add a comment explaining why `'assistant'` is metrics-only.

BogLesh: I've aligned turn-detector.ts and GeminiSessionAdapter to treat type: 'assistant' as a first-class assistant message alongside 'gemini'. Added unit test coverage in turn-detector.test.ts to ensure both processors handle turns in lockstep. 
Thanks

src/utils/pricing.json
BogLesh: Makes sense, thanks for catching that. 
Dropped the bare "gemini" key in favor of explicit versioned keys (e.g., gemini-3-7-flash, gemini-3-5-flash) and updated the unit tests accordingly.

yanaSelin (approved):

PR #537, opened 2026-09-05
fix(assistants): improve error logging in chat to show actual error details

gokhanozdemir:
## Summary
Errors thrown during assistants chat were serialized as `[object Object]`, hiding the actual message, name, code, and stack trace needed for debugging, and could be mistaken for an authentication error.

Root cause: `logger.error(message, error)` only unwraps `.message`/`.stack` when the second argument is an actual `Error` instance. Any plain object — including the structured context produced by `createErrorContext()` — fell through to `String(error)`, which yields `[object Object]`.

Fixes #499

## Changes
- `src/utils/logger.ts`: added `formatNonErrorValue()` and used it in `Logger.error()`'s non-`Error` branch (both the file-log and console-log paths) so plain objects are `JSON.stringify`'d instead of `String()`-ed. This fixes the bug at its source for every caller, including the existing `logger.error('Assistant chat API call failed', context)` call in `handleChatError`.

## Impact
Error output during assistants chat failures (and any other `logger.error()` call site passing a non-`Error` object) now shows actionable detail instead of `[object Object]`; no change to normal (non-error) behavior.

## Checklist
- [x] Self-reviewed
- [x] Manual testing performed
- [ ] Documentation updated (if needed)
- [x] No breaking changes (or clearly documented)

PR #538, opened 2026-09-05
fix(agents): add file_not_found correlation status for missing transcripts

### LINKED ISSUE #523: Interactive Claude sessions: transcript not persisted at reported path → session metrics silently empty (0 records)
## Description

For interactive Claude Code sessions launched via `codemie-claude`, Claude Code (2.1.218) does not write its session transcript JSONL to `~/.claude/projects/<cwd>/<session>.jsonl` — the project dir contains only a `memory/` subdir. However, Claude's SessionStart/UserPromptSubmit hook payload still reports that `transcript_path`, and CodeMie's correlation marks it `status: "matched"`. At `Stop`/`SessionEnd`, the metrics processor reads the absent path, reports **"0 records processed"**, and writes no `sessions/*_metrics.jsonl`. Session metrics are silently lost, and correlation reports `matched` for a file that doesn't exist — masking the gap.

Headless `--task` mode persists the transcript normally, so its metrics work.

## Steps to Reproduce

1. Run the agent suite: `npm run test:integration:agent` (or just TC-024 in `tests/integration/agent-model.test.ts`).
2. TC-024 drives `codemie-claude` interactively via node-pty, does a `/model` switch + a prompt, then `/exit`.
3. Observe the session's `logs/debug-*.log`: hooks fire, but both `Stop` and `SessionEnd` log `Session processing complete: 0 records processed`.
4. `~/.claude/projects/<cwd-hash>/` contains only `memory/` — no `.jsonl`.

## Expected Behavior

Interactive-session metrics are recorded (a `sessions/*_metrics.jsonl` is written with the model used), as with `--task` mode.

## Actual Behavior

No metrics file is produced; `getLatestMetricsRecord` throws `No metrics files found`. The transcript at the correlated path never exists on disk.

## Investigation (ruled out)

Temp vs persistent cwd · workspace trust · the `/model` switch · provider (`ai-run-sso` vs `anthropic-subscription`) · session duration / turn count (4 turns / 75s) · flush-on-clean-exit. A "normally" launched interactive `codemie-claude` session (same `claude --plugin-dir …` invocation) *does* persist its transcript, so the trigger appears specific to how a programmatically-driven (node-pty) interactive session is handled by Claude Code — i.e. likely an upstream transcript-persistence behavior, not CodeMie logic.

## Partial mitigation (already landed)

`src/cli/commands/hook.ts` now logs an explicit `WARN` when the reported transcript path is missing (or empty) at `Stop`/`SessionEnd`, instead of silently reporting "0 records processed", so the gap is diagnosable.

## Suggested follow-ups

- Don't mark correlation `matched` for a transcript path that doesn't exist on disk.
- Bisect Claude Code versions to confirm where interactive-PTY transcript persistence changed.
- Adjust TC-024 to assert against the API-sent session metric, or drive the model check via `--task`.

## Environment

| Field | Value |
|---|---|
| OS | Darwin 25.5.0 arm64 |
| Node.js | v26.5.0 |
| CodeMie CLI | 0.15.0 |
| Claude Code | 2.1.218 |
| Shell / Terminal | zsh / ghostty |

gokhanozdemir:
## Summary
Interactive Claude sessions can fail to persist transcripts at their expected file paths. PTY-driven sessions weren't validated for file existence, so the correlation logic incorrectly marked non-existent files as "matched," producing empty session metrics. This adds a `file_not_found` correlation status so missing transcripts are handled gracefully instead.

Fixes #523

## Changes
- Add a `file_not_found` status to the correlation type system
- Validate file existence before marking correlations as matched in the hook command
- Update the session synchronizer to skip gracefully when files are missing, instead of failing outright

## Impact
No user-facing behavior change for existing successful correlations; missing-transcript cases now report a clear status instead of silently producing empty metrics.

## Checklist
- [x] Self-reviewed


codemie-ai:
[AUTO_CLOSE_WARNING] ⏰ This pull request has had no activity for 7 days (last activity: 2026-09-05).

It will be automatically closed no earlier than **2026-09-19 (UTC)** (~6 more days).

Any activity from anyone — a comment, a push, a review, reopening — restarts the countdown from that moment. This warning will be deleted automatically once that happens.

gokhanozdemir:
## Adversarial review

**Critical regression:** `event.transcript_path ? existsSync(...) : false` forces `correlationStatus` to `'file_not_found'` whenever `transcript_path` is empty. But `codemie-code`, `codex`, `copilot-cli`, `opencode`, and `pi` all fire `SessionStart` with `transcript_path: ''` by design (path not known yet):

- `src/agents/plugins/codemie-code.plugin.ts:233`
- `src/agents/plugins/codex/codex.plugin.ts:191`
- `src/agents/plugins/copilot-cli/copilot-cli.plugin.ts:161`
- `src/agents/plugins/opencode/opencode.plugin.ts:210`
- `src/agents/plugins/pi/pi.plugin.ts:376`

Nothing anywhere ever rewrites `correlation.status` after creation (verified repo-wide), so these sessions get stuck at `file_not_found` permanently, and `SessionSyncer.sync()` silently (`success: true`) skips them forever. This breaks metrics/conversation sync for most agents, not just interactive Claude PTY sessions — a much bigger regression than the bug being fixed.

**Broken pre-existing test:** `tests/integration/agent-model.test.ts:329` skip-guards on `correlationStatus === 'failed'`, written specifically to tolerate this exact issue (#523). The new code never produces `'failed'` here — it produces `'file_not_found'` — so the guard is dead and the test will fail instead of skipping gracefully.

**Other issues:**
- Race on re-entrant `SessionStart` (e.g. `compact`) can downgrade an already-`matched` live session to `file_not_found` with no recovery path.
- One-shot `existsSync` check at record-creation time doesn't cover a transcript appearing moments later, or one deleted/rotated after being marked `matched`.
- Duplicated existsSync/ternary/warn block across two branches in `hook.ts`; redundant dynamic `import('node:fs')` for a builtin already statically imported elsewhere.

**Fix applied in a follow-up push:** only mark `file_not_found` when a non-empty `transcript_path` was reported and is missing; keep prior behavior when no path was reported (the deliberate "not known yet" case); dedupe the two branches; use a static `existsSync` import; update the test's skip guard to also accept `file_not_found`.

gokhanozdemir:
Cherry-picked the Windows CI teardown fix from #537 (`110add2f`) onto this branch to unblock the flaky `migration-runner-ordering.test.ts` failure we hit here.

That commit matches an existing established pattern already used in several test files in this codebase (`claude.metrics-processor-names.test.ts:104`, `claude.metrics-processor-clear.test.ts:90`, `codex-models.test.ts:141`, `metrics-upload-contract.test.ts:112`): wrap `rmSync(..., { recursive: true, force: true })` with `maxRetries: 5, retryDelay: 100`, plus a try/catch that swallows the cleanup error. Reason: on Windows, `rmSync`'s `force` option only suppresses `ENOENT`, not the `ENOTEMPTY`/`EBUSY` that can occur when a file handle (e.g. from a logger stream) is released a moment after the recursive delete already removed the children.

Not a hack — it's the codebase's already-established idiom for this class of flaky Windows teardown, applied here to the migration runner test that was failing on our Windows CI job.

codemie-ai:
[AUTO_CLOSE_WARNING] ⏰ This pull request has had no activity for 7 days (last activity: 2026-09-15).

It will be automatically closed no earlier than **2026-09-29 (UTC)** (~6 more days).

Any activity from anyone — a comment, a push, a review, reopening — restarts the countdown from that moment. This warning will be deleted automatically once that happens.

PR #551, opened 2026-09-10
fix(config): add migration 008 to fix backslash paths in installed hooks

oleksii-hryshyn:
## Summary
Migration 008 heals already-installed Claude and Gemini hooks.json files that contain Windows backslash paths written by migration 006. Bash (Git Bash / WSL) silently consumes `\X` sequences as escape codes, so `C:\Users\...\codemie.cmd` becomes `C:Users...codemie.cmd` at runtime, causing every hook to fail with `command not found`.

## Changes
- `src/utils/hook-command.ts`: export new `fixCommandTreeSlashes(node)` that walks the hooks tree and applies `toForwardSlash()` to any `command` field containing a backslash
- `src/migrations/008-fix-windows-hook-path-slashes.migration.ts`: new one-shot migration that reads both Claude and Gemini hooks.json, calls `fixCommandTreeSlashes`, and writes back only if something changed; returns `{ success: false }` on write failure so transient errors are retried
- `src/migrations/index.ts`: registers migration 008

## Why a migration instead of relying on `codemie install claude`
Migration 006 already ran and is marked applied for affected users, so it will not re-run. Even if it did, `rewriteHooksCommandTree` only rewrites commands starting with `codemie` — not already-absolute backslash paths. And the plugin version was not bumped in the fix release, so `installer.install()` sees `already_exists` and skips `localizeInstalledHooks`.

## Testing
- [ ] Build passes (`npm run build`) ✅
- [ ] Lint clean, typecheck clean, secrets scan clean ✅ (all pre-commit hooks passed)
- [ ] Manual: on a Windows machine with a backslash-path hooks.json, first `codemie` run after update applies the migration and hooks execute correctly

## Checklist

Refs: EPMCDME-14762


8nevil8:
we should update plugin version and bring hooks there. it suppose to replace existing hooks. migration is not applicable here

PR #555, opened 2026-09-13
fix(proxy): route VS Code through selected profile model

nadavsinai-philips:
Makes proxy connect --vscode and --vscode-insiders register the active profile's model ID when it maps to a supported VS Code capability template. This prevents a provider-qualified CodeMie ID such as openai.gpt-5.6-sol from being overwritten with the incompatible dated catalog ID. Retains the static catalog fallback for profile-less or unrecognized models. Also makes Node types explicit for stable typecheck. Validated by the required hook: lint, 319 focused proxy tests, typecheck, and secret scan.


8nevil8:
@nadavsinai-philips , could you pls fix CI and we'll proceed with review?

8nevil8:
@nadavsinai-philips could you pls rebase? we'll proceed with merge after that

PR #558, opened 2026-09-15
fix(utils): ignore a `codemie` on PATH owned by another npm package

8nevil8:
## Summary

When the Claude plugin was installed from inside an npm script, the hook installer wrote opencode's binary into the Claude hooks instead of the CodeMie CLI. `resolveCodemieBinary()` took whatever `which codemie` returned. In npm scripts `node_modules/.bin` comes first on PATH, and `@codemieai/codemie-opencode` (<= 0.0.47) links its binary there as `codemie`. After that, `codemie hook` silently did nothing, so no session tracking or sync ran, and `codemie sound Stop` exited 1. Claude Code showed that as `Stop hook error: Failed with non-blocking status code: No stderr output`.

## Changes

- `resolveCodemieBinary()` now skips a PATH result whose real path is inside an installed npm package other than `@codemieai/code`, and falls through to the existing `argv[1]` fallback.
- Paths outside `node_modules` (npm link, standalone installs, Windows `.cmd` shims) and paths that can't be read are handled exactly as before.
- Two unit tests use a real `node_modules/.bin/codemie` symlink fixture: one where the link belongs to another package and is rejected, one where it belongs to `@codemieai/code` and is kept. On Windows the fixture uses the direct path, because symlinks need elevated rights there.

Related: codemie-ai/codemie-opencode#25 drops the wrapper's `codemie` bin entirely. That fixes the source, and this PR guards against the same problem from any other package. The `@codemieai/codemie-opencode` dependency bump will follow once that is published.

## Impact

Before: `~/.codemie/claude-plugin/hooks/hooks.json` → `…/codemie-code/node_modules/.bin/codemie hook`, which is opencode.
After: the installer skips that link. Users with an already-broken `hooks.json` recover on the next plugin version bump, or by running `codemie install` again from a normal shell.

## Checklist

- [x] Self-reviewed
- [x] Manual testing performed: with `node_modules/.bin` prepended to PATH, `which codemie` returns the opencode link, and `resolveCodemieBinary()` now rejects it and falls back to the running entry. The affected Stop/`sound` hooks exit 0 once they point at the real CLI.
- [x] `hook-command`, `BaseExtensionInstaller.hooks`, `006-resolve-hook-command-paths` and `codemie-code.hooks` unit tests pass (20/20). eslint and `tsc --noEmit` are clean.
- [ ] Documentation updated (if needed): not needed
- [x] No breaking changes


8nevil8:
Superseded by #561 (merged).

That PR fixes the same root cause in resolveCodemieBinary() — it rejects PATH shims resolving into @codemieai/codemie-opencode — and goes further: already-broken hooks.json installs are repaired automatically via the plugin version bump (1.0.27 / 1.0.3) through localizeInstalledHooks() -> resolveHookCommand(), which now rewrites absolute shim paths. The upstream fix codemie-ai/codemie-opencode#25 also removes the rogue codemie bin at the source.

What this PR still has that #561 does not: generic any-foreign-package detection (package.json walk-up) and two unit tests with symlink fixtures. If we want those, they'll be easier to re-apply as a small follow-up on top of current main than to rebase this branch through the now-conflicting resolveCodemieBinary(). Closing.

PR #563, opened 2026-09-17
feat(cli): feature tips catalog, session tips, and whats-new release notes

8nevil8:
## Summary
Users can't keep up with CodeMie Code's growing capability set — features ship, but there is no mechanism that tells users they exist. This adds a curated, easy-to-maintain feature-tips system surfaced at session start/end and across high-value CLI surfaces, plus a What's New flow driven by a shipped CHANGELOG.md.

## Changes
- **Tips catalog** (`src/utils/tips.ts`): 75 human-toned tips covering every command *and* meaningful subcommand/flag (e.g. `codemie docs install codegraph`), verified against live `--help` output; no-repeat rotation state; `CODEMIE_TIPS=0` opt-out; theme-proof highlight rendering (explicit yellow/blue backgrounds — readable on dark and light terminals)
- **Session surfaces** (`BaseAgentAdapter`): one tip at session start and at both exit paths, inside the existing `silentMode` gates (ACP/`--task` unaffected)
- **CLI surfaces**: tips after `codemie doctor`, on first-run welcome, quick start, and post-setup screens
- **`codemie tips`**: browsable grouped catalog (`--random`, `--category`); validates tip command references against the live commander tree and drops stale tips automatically — retiring a command can't leave a dangling tip
- **What's New**: `CHANGELOG.md` (Keep a Changelog, seeded 0.14.0–0.15.1, shipped in the npm package) + `codemie whatsnew [--all] [--version <v>]` + a once-per-upgrade startup notice; `release.sh` warns when the changelog lacks a section for the version being released
- **Commander fix**: `enablePositionalOptions()` so subcommand flags like `--version` are no longer intercepted by the root program
- Design spec: `docs/superpowers/specs/2026-09-17-feature-tips-design.md`

## Impact
Before: session start/end showed personality-only messages; no feature discovery mechanism existed. After: every session teaches one capability, `codemie tips` browses all 75, and users see what changed after each upgrade.

## Testing
- [ ] Tests added/updated
- [x] Manual testing done (build, typecheck, lint zero-warning; `tips`/`whatsnew` variants, category filter, rotation cycle, upgrade notice once-per-version, `CODEMIE_TIPS=0` suppression, silent-mode gating)

## Checklist


homich1991 (approved):

homich1991:
## Code Review Findings

### Bugs

**1. `src/cli/index.ts:67`** — `.enablePositionalOptions()` on the root program breaks recognition of root-level options once a subcommand name appears in argv.
- Failure scenario: `codemie list --version`, `codemie install --version`, `codemie proxy --version` (root option placed after the subcommand — previously valid) now fail with `error: unknown option '--version'` instead of printing the version.

**2. `bin/codemie.js:53`** — The once-per-upgrade "What's new" notice fires on a brand-new install, because a missing `.last-seen-version` marker is treated the same as "upgraded from a different version."
- Failure scenario: `getLastSeenVersion()` returns `null` on a fresh `~/.codemie`. `null !== currentVersion` is `true`, so the very first `codemie` invocation ever run prints the "Updated to X.Y.Z" banner, which is misleading for a new install — and it runs before `FirstTimeExperience.showWelcomeMessage()`.

**3. `bin/codemie.js:38-65`** — The whatsnew notice block runs unconditionally on every CLI invocation (only gated by `isTestEnvironment`), with no check for non-interactive/scripted use.
- Failure scenario: right after an upgrade, `codemie --version`, `codemie doctor`, or a CI script invoking `codemie <cmd>` prints the changelog banner to stdout before the command's own output — can corrupt scripted/piped consumption on that one run.

**4. `CHANGELOG.md`** — `getReleaseNotes(currentVersion)` requires an exact `## [x.y.z]` heading match, but this PR's entries are filed under `## [Unreleased]`; `scripts/release.sh` only warns (non-blocking) if the heading is missing at release time.
- Failure scenario: if a release is cut without renaming `[Unreleased]` to `[<version>]`, `getReleaseNotes(currentVersion)` returns `[]` on first launch after upgrade — no notice prints, but `markVersionSeen(currentVersion)` still fires, so that version's notice is silently and permanently lost.

### Duplication / Quality

**5. `src/agents/core/BaseAgentAdapter.ts` (~590-596, 759-764, 934-939)** — The "goodbye message + end tip + Powered by" block is duplicated verbatim across independent exit paths (built-in-handler path and spawned-CLI exit path, plus the start-tip block).
- Risk: a future change to tip-rendering logic applied to one path and not the other leaves session-end output inconsistent depending on code path — easy to miss since the blocks are ~180 lines apart.

**6. `src/utils/tips.ts:112-123`** — `isTipsEnabled()` is a verbatim copy of `isAutoUpdateEnabled()` in `src/utils/cli-updater.ts:122-133` (same default/parsing logic), with no shared `parseBoolEnv()` helper.
- Risk: a fix or extension to `CODEMIE_*` boolean parsing has to be applied in both copies by hand; nothing signals they must stay in sync.

**7. `src/utils/tips.ts:560-584`** — `getSessionTip()` performs a full read-modify-write of `~/.codemie/.tips-state.json` at both session start and session end, with no locking.
- Failure scenario: two `codemie` processes running concurrently against the same home directory (e.g. parallel agent sessions) can race on the read-modify-write, dropping or duplicating rotation entries in `.tips-state.json`.

---
*Posted via automated code review (`/code-review`).*

8nevil8:
All 7 findings addressed in the latest commits — thanks for the thorough review.

**1. `enablePositionalOptions` regression** — fixed differently: the option is removed entirely and `codemie whatsnew` now takes a positional `[version]` arg (`codemie whatsnew 0.14.0`) instead of `--version <v>`. `codemie list --version` prints the root version again; no commander behavior change ships. (6e9fe47)

**2. Fresh-install banner** — fixed: a missing `.last-seen-version` marker now just records the version silently; the "Updated to" banner only renders when a previous marker exists and differs. (6e9fe47)

**3. Scripted-use corruption** — the notice is now skipped when stdout is not a TTY and when argv contains `--version`/`-V`/`--task`/`--help`, in addition to the existing test-env skip. (6e9fe47)

**4. `[Unreleased]` mismatch losing the notice** — `getReleaseNotes(version)` now falls back to the newest versioned entry when the exact heading is missing (debug-logged), so a release cut without renaming `[Unreleased]` still shows users the latest notes. (6e9fe47)

**5. Duplicated session-end blocks** — extracted into a single `renderSessionEnd()` used by both exit paths; behavior unchanged. (c968a54)

**6. Duplicated boolean env parsing** — new `parseBooleanEnv()` in `src/utils/env.ts`, now used by both `isTipsEnabled()` and `isAutoUpdateEnabled()`. (64c198c)

**7. Rotation read-modify-write race** — the state file is now written atomically (tmp + rename), eliminating torn writes. Full locking deliberately skipped: the residual worst case is a repeated tip, which is harmless — noted in the function docs. (64c198c)

